PR #9, opened 2024-09-04
Stan and PyRenew model comparison doc

damonbayer:
Doc can be rendered by executing `poetry run quarto render notebooks/model_comp.qmd` from within the `pyrenew-covid-wastewater` directory.


notebooks/data/fit/stan_data.json
AFg6K7h4fhy2, line 1: Would be good if these value choices are documented somewhere. I've "viewed" this in the sense that I trust the change in this file is appropriate.

notebooks/model_comp.qmd
AFg6K7h4fhy2, line 1: Deferring to others (e.g. DHM) on this, which ofc constitutes the bulk of this PR. This note can be resolved when read. Including so DB doesn't perceive my non-response as wait time.

AFg6K7h4fhy2 (commented):
Deferred my review to others for comparison section. The other changes (minor) look good.

notebooks/data/fit/stan_data.json
damonbayer: This is a result of the change to `notebooks/wwinference.Rmd`. The stan and PyRenew models are slightly different because the stan model is hierarchical with one "site," while the PyRenew model is not hierarchical. This/that change makes it so that the "site" Rt behaves very very similarly to the "global" Rt.

sbidari: The site-level model would be hierarchical, so this change is only necessary in `notebooks/data/fit_hosp_only/stan_data.json`?

damonbayer: Good catch. Addressed in https://github.com/CDCgov/pyrenew-covid-wastewater/pull/9/commits/58a0d82f34412710a94bd25d20d0aaa07424f16e.

sbidari (changes requested):
You need to run the `wwinference.Rmd` before the model comparison demo renders as it loads `.csv` files which are not tracked. I would add a note at the beginning of the `model_comp.qmd` to indicate this.

notebooks/model_comp.qmd
```diff
@@ -0,0 +1,158 @@
...
+  ggtitle("Vignette Data Model Comparison") +
+  scale_y_continuous("Hospital Admissions") +
+  scale_x_continuous("Time") +
+  theme(legend.position = "bottom")
+```
+
+
+
+## Rt Comparions
+
```
sbidari, line 119: Maybe add a note here to explain why the two Rt estimates are different. Due to the lower prior variance for Rt in pyrenew model, right?

damonbayer: The Rt comparisons should not be different. You are probably seeing results from a previous Stan fit and may need to rerun the Stan model.

damonbayer: Perhaps, but I'm not completely sure. They seem to basically be in agreement after `t=0`
unnamed-chunk-5-1

damonbayer:
> You need to run the `wwinference.Rmd` before the model comparison demo renders as the demo uses `.csv` files which are not tracked. I would add a note at the beginning of the `model_comp.qmd` to indicate this.

addressed in https://github.com/CDCgov/pyrenew-covid-wastewater/pull/9/commits/3ec45c17a1b1c27b1a4d2574bc3a11ab8286702c

sbidari (approved):

sbidari:
Looks good to me!

PR #16, opened 2024-09-13
NSSP Demo

damonbayer:
Looking for thoughts on:

- Format for storing priors / model params
  - I am using a `priors.py` file which constructs the necessary `RandomVariable` objects. This is very convenient for running the code and extremely precise in its meaning. However, it feels wrong to me.
- Location of "private data"
  - I just used a `gitignore`'d file for now.

Don't worry too much about the "pipeline" bits of this PR. These need to azure-ified anyway.


dylanhmorris (approved):
LGTM

PR #49, opened 2024-10-11
Wastewater Model

sbidari:
This PR replicates the WWW model as implemented [here](https://github.com/cdcgov/ww-inference-model/). 

---------------fixed-----------------------------------------------
There is one issue with using predictive methods, specifically for the observed hospital admissions. Currently, the posterior predictive distribution for `observed_hospital_admissions` does not extend beyond the available datapoints because of how it is defined in the model. This was done to account for possible differences in the length of reporting time for wastewater and hospital admissions data, following the approach in the stan code. I am not sure how to circumvent this issue for the pyrenew model and would love to hear ideas/suggestions. 
@dylanhmorris @damonbayer @kaitejohnson


kaitejohnson (commented):
This is looking really great @sbidari! I am following like 95% of the model which is shocking considering I don't know anything about numpyro :) 

My main comment is that I don't see anywhere where forecasts quantities would be generated. I think that's okay, because my guess is that they way that you'll be implementing this is doing inference by sampling the model you wrote, getting the posterior parameters, and then using those to do forecasting? 

Which I think will require a step where you do that forward simulation of the R(t), infections, and observations. I think you can do this for both the calibration and nowcast/forecast periods, but you might already have gotten the calibration period "for free" from the sampling. 

Others who know more numpyro feel free to correct me if I'm wrong!

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+                    len(data_observed_hospital_admissions),
+                    self.max_ww_sampled_days,
+                )
+        else:
+            if (
+                data_observed_hospital_admissions is not None
+                or data_observed_log_conc is not None
+            ):
+                raise ValueError(
+                    "Cannot pass both n_datapoints and "
```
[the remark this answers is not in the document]
damonbayer: To do forecasting, you pass the number of datapoints, rather than the data itself.

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+                    "data_observed_hospital_admissions "
+                    "or data_observed_log_conc"
+                )
+            else:
+                n_datapoints = n_datapoints
+
+        n_weeks_post_init = n_datapoints // 7 + 1
```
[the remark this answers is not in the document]
damonbayer: Ceiling and casting to an int causes issues with Jax.

I think -((-x) // y) would make things correct, but still only use integer division.

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+                "clipped_initial_exp_growth_rate_non_ref_subpop",
+                DistributionalVariable(
+                    "initial_exp_growth_rate_non_ref_subpop_raw",
+                    dist.Normal(
+                        mean_initial_exp_growth_rate,
+                        sigma_initial_exp_growth_rate,
+                    ),
+                    reparam=LocScaleReparam(0),
+                ),
+                transforms=lambda x: jnp.clip(x, -0.01, 0.01),
```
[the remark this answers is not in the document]
damonbayer: Typically, you would use a [`TruncatedDistribution`](https://num.pyro.ai/en/stable/distributions.html#truncateddistribution). If this is not applicable here, we should leave a note in the comments.

damonbayer:
> My main comment is that I don't see anywhere where forecasts quantities would be generated. I think that's okay, because my guess is that they way that you'll be implementing this is doing inference by sampling the model you wrote, getting the posterior parameters, and then using those to do forecasting?

The forecast quantities are generated here:
https://github.com/CDCgov/pyrenew-hew/pull/49/files#diff-a74ba37ba755f355d4f4246e8843051cbf8e69194ffed847025d7f37dffe2ad9R502-R504

pyrenew_hew/site_level_dynamics_model.py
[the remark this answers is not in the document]
kaitejohnson: The total number of data points you want in the calibration + nowcast/forecast period? 

This would still be when you are using the data to calibrate the model, correct? It's just if you want to produce estimates that extend beyond the time range of the data, you would need to tell the model the total number of time points you want to generate estimates for?

kaitejohnson:
> > My main comment is that I don't see anywhere where forecasts quantities would be generated. I think that's okay, because my guess is that they way that you'll be implementing this is doing inference by sampling the model you wrote, getting the posterior parameters, and then using those to do forecasting?
> 
> The forecast quantities are generated here: https://github.com/CDCgov/pyrenew-hew/pull/49/files#diff-a74ba37ba755f355d4f4246e8843051cbf8e69194ffed847025d7f37dffe2ad9R502-R504
> 
> When you extend the length of the data beyond what was used in fitting, everything "just works". You don't have to do anything special to distinguish the model fitting period and the forecasting period. When sampling from the posterior predictive `numpyro` uses the posterior samples when it can, and simulates "future" posterior samples (for example to extend Rt into the future) conditional on the existing posterior samples.

Could you resend that link for some reason it isn't pointing to a specific place for me.

Ok so this is the same as the stan model. Then I guess I am confused what isn't "just working" with extending the `observed_hospital_admissions` beyond the number of data points in the data as @sbidari described!

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+                    "Either n_datapoints or data_observed_hosp_admissions "
+                    "must be passed."
+                )
+            elif (
+                data_observed_hospital_admissions is not None
+                and data_observed_log_conc is None
+            ):
+                n_datapoints = len(data_observed_hospital_admissions)
+            else:
+                n_datapoints = max(
```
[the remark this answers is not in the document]
sbidari: Yes!

pyrenew_hew/site_level_dynamics_model.py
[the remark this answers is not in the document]
sbidari: `LocScaleReparam` does not support Truncated distributions.

pyrenew_hew/site_level_dynamics_model.py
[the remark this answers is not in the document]
sbidari: Right! You pass the number of datapoints when calling predictive methods on the model after it has been calibrated.
Or if you just want to run forward simulations form the model, you pass the number of datapoints.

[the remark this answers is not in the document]
damonbayer: You cannot comforting and forecasting at the same time. You can fit by passing the actual data. You can create a (prior or posterior) predictive distribution by passing the number of time points (which can be arbitrarily few or many).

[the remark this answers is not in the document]
kaitejohnson: Ok I interpreted your "it just works" statement incorrectly then. 

To simulate from the posterior predictive, do you pass in a fitted model object and then the total number of time points you want to generate posterior predictive observations for (e.g. calibration + nowcast/forecast)? I'm just struggling to see the issue with generating data beyond the number of days of the observed hospital admissions that @sbidari described in the PR description.

sbidari:
> Ok so this is the same as the stan model. Then I guess I am confused what isn't "just working" with extending the `observed_hospital_admissions` beyond the number of data points in the data as @sbidari described!

The `observed_hospital_admissions` variable is not extended beyond the number of data points when using predictive method. I think the reason is how this variable is defined in the model - the mean/mu for the hospital admissions random variable only receives length equal to the hospital admissions data:
```
observed_hospital_admissions = hospital_admission_obs_rv(
            mu=latent_hospital_admissions[self.hosp_times],
            obs=data_observed_hospital_admissions,
        )
```
I believe this is preventing this variable form being `just extended` to the number of datapoints passed.

notebooks/site_level_dynamics_model_demo.qmd
damonbayer, line 1: @sbidari Please share a rendered version of this document on Teams

damonbayer:
> The `observed_hospital_admissions` variable is not extended beyond the number of data points when using predictive method. I think the reason is how this variable is defined in the model - the mean/mu for the hospital admissions random variable only receives length equal to the hospital admissions data:
> 
> ```
> observed_hospital_admissions = hospital_admission_obs_rv(
>             mu=latent_hospital_admissions[self.hosp_times],
>             obs=data_observed_hospital_admissions,
>         )
> ```
> 
> I believe this implementation is preventing `observed_hospital_admissions` variable form being "just extended" to the number of datapoints passed.

Why does this same problem not affect the current model, which features a similar code block?

https://github.com/CDCgov/pyrenew-hew/blob/main/pyrenew_hew/hosp_only_ww_model.py#L226-L229

notebooks/data/fit/stan_data.json
damonbayer, line 1: Why is this file changed?

notebooks/wwinference.Rmd
damonbayer, line 1: Are the lines 
```r
params$sigma_rt_prior <- 0.0001 # This model contains a single "site," so we
# decrease the prior variance to make the "site" Rt nearly the same as the
# "global" Rt.
```

still relevant? I think this was changed in the most recent version on the R package.

pyrenew_hew/initialization.py
damonbayer, line 1: Maybe make an issue to improve the model such that this isn't needed.

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+        ww_ml_produced_per_day=None,
+        ww_uncensored=None,
+        ww_censored=None,
+        ww_sampled_lab_sites=None,
+        ww_sampled_subpops=None,
+        ww_sampled_times=None,
+        ww_log_lod=None,
+        lab_site_to_subpop_map=None,
+        max_ww_sampled_days=None,
+        include_ww=0,
```
damonbayer, line 72:
```suggestion
        include_ww=False,
```

pyrenew_hew/site_level_dynamics_model.py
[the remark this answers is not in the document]
damonbayer: This should probably just be a Normal with a very tiny sd. Can be a separate issue.

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+                [
+                    log_rtu_ref_subpop_in_week[:, jnp.newaxis],
+                    log_rtu_non_ref_subpop_in_week,
+                ],
+                axis=1,
+            )
+        else:
+            i_first_obs_over_n_subpop = i_first_obs_over_n_ref_subpop
+            initial_exp_growth_rate_subpop = initial_exp_growth_rate_ref_subpop
+            log_rtu_subpop_in_week = log_rtu_ref_subpop_in_week[:, jnp.newaxis]
```
damonbayer, line 314: Mild preference for reversing the order of the if else here. (If <= 1 subpop, do the simple thing. Else do the complex thing) By the time I got to this, I forgot what the `if` was checking.

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+        hospital_admission_obs_rv = NegativeBinomialObservation(
+            "observed_hospital_admissions", concentration_rv=self.phi_rv
+        )
+        observed_hospital_admissions = hospital_admission_obs_rv(
+            mu=latent_hospital_admissions[self.hosp_times],
+            obs=data_observed_hospital_admissions,
+        )
+
+        # wastewater component
+        if self.include_ww:
```
damonbayer, line 445: Shouldn't all of the subpopulation stuff be put under a similar `if self.include_ww:` check?

damonbayer (commented):
@sbidari Took a brief look. Will review in more depth once the fix for the posterior predictive is in.

pyrenew_hew/site_level_dynamics_model.py
sbidari: changed in https://github.com/CDCgov/pyrenew-hew/pull/49/commits/259e895ce55a0248dca4ffd590dc418dfb172dc9

pyrenew_hew/site_level_dynamics_model.py
sbidari: Good point! changed in https://github.com/CDCgov/pyrenew-hew/pull/49/commits/259e895ce55a0248dca4ffd590dc418dfb172dc9

pyrenew_hew/initialization.py
sbidari: The initialization values are not needed any more. Will remove this file

notebooks/wwinference.Rmd
sbidari: I am not sure where these code blocks appear

pyrenew_hew/site_level_dynamics_model.py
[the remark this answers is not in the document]
sbidari: Changed the code to use `-((-x) // y)` in https://github.com/CDCgov/pyrenew-hew/pull/49/commits/cc00bd9aaa4abb1c5561b99fdf1844c72f6cf78d

pyrenew_hew/site_level_dynamics_model.py
```diff
@@ -0,0 +1,566 @@
...
+                - jnp.log(self.ww_ml_produced_per_day)
+            )
+            numpyro.deterministic("state_log_c", state_log_c)
+
+            expected_state_ww_conc = jnp.exp(state_log_c)
+            numpyro.deterministic(
+                "expected_state_ww_conc", expected_state_ww_conc
+            )
+
+            state_rt = (
```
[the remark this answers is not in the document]
sbidari: I am considering these comments/questions as resolved but let me know if anything was not addressed.

notebooks/wwinference.Rmd
damonbayer: https://github.com/CDCgov/pyrenew-hew/blob/cc00bd9aaa4abb1c5561b99fdf1844c72f6cf78d/notebooks/wwinference.Rmd#L484-L486

sbidari: Right! These are not necessary. 
Also realized I can make some arguments as optional for running hosp only model. Made these changes in https://github.com/CDCgov/pyrenew-hew/pull/49/commits/595dbb8578de1ff46d46cd1d60571be2235fa314

pyrenew_hew/site_level_dynamics_model.py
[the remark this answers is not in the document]
sbidari: Confirming non-Truncated normal works here.

dylanhmorris:
> Why does this same problem not affect the current model, which features a similar code block?
> 
> https://github.com/CDCgov/pyrenew-hew/blob/main/pyrenew_hew/hosp_only_ww_model.py#L226-L229

This version of the model masks the `mu` to allow for there to be dates during the fitting period without observed admissions values. That prevents the "set `data_observed_hospital_admissions=None` and and extend `n`" approach from just working. I proposed a somewhat inelegant solution to @sbidari yesterday (which worked), but it occurs to me here that a much more readable one is having a `is_predictive` argument to the `Model.sample()` call and then doing:

```python
if not is_predictive:
   mu_obs_hosp = latent_hospital_admissions[self.hosp_times]
else:
  mu_obs_hosp = latent_hospital_admissions

observed_hospital_admissions = hospital_admission_obs_rv(
            mu=mu_obs_hosp
            obs=data_observed_hospital_admissions,
        )
```

@sbidari note that this will also avoid any issue with attempting to sample an unobserved discrete site via MCMC.

.gitignore
damonbayer, line 1: I think these changes could be reversed?

sbidari: reverted to previous state in https://github.com/CDCgov/pyrenew-hew/pull/49/commits/a20309e4c5fd537a79cabb4dc7f3a33d3c712db5

notebooks/data/fit_hosp_only/stan_data.json
damonbayer, line 1: I think these changes are compatible with the `hosp_only_ww_model`, can you rerun `hosp_only_ww_model.qmd` to verify?

pyrenew_hew/plotting.py
damonbayer, line 1: Unclear exactly what the fix is, but is there a way to have the different `Posterior r_subpop_t` plots have different titles to indicate which subpop is being plotted?

notebooks/data/fit_hosp_only/stan_data.json
sbidari: Confirming changes compatible with `hosp_only_ww_model`

pyrenew_hew/plotting.py
sbidari: Yeah, that would be nice. I will create an issue for creating plots of wastewater concentrations and Rt for different subpopulations.

demos/ww_model/data/fit/stan_data.json
damonbayer, line 1: I believe this is out of date compared to https://github.com/CDCgov/pyrenew-hew/pull/186/files, which uses the latest data from https://github.com/CDCgov/ww-inference-model as of last week.

damonbayer:
@sbidari Can you share the rendered qmd file somewhere for me to look at?

pyrenew_hew/utils.py
```diff
@@ -1,9 +1,54 @@
...
-        np.power(mean, 2) / np.sqrt(np.power(sd, 2) + np.power(mean, 2))
+    logmean = jnp.log(
+        jnp.power(mean, 2) / jnp.sqrt(jnp.power(sd, 2) + jnp.power(mean, 2))
     )
-    logsd = np.sqrt(np.log(1 + (np.power(sd, 2) / np.power(mean, 2))))
+    logsd = jnp.sqrt(jnp.log(1 + (jnp.power(sd, 2) / jnp.power(mean, 2))))
     return logmean, logsd
+
+
+def normed_shedding_cdf(
```
damonbayer, line 12: A docstring would be very helpful here.

damonbayer, line 12: more descriptive names (or more comments) would be very helpful for understanding this code.

damonbayer:
@sbidari I think we should also include an equivalent of [`demos/hosp_only_ww_model/model_comp.qmd`](https://github.com/CDCgov/pyrenew-hew/blob/main/demos/hosp_only_ww_model/model_comp.qmd) so we can easily compare results to the stan model.

pyrenew_hew/ww_site_level_dynamics_model.py
```diff
@@ -0,0 +1,570 @@
...
+        offset_ref_log_r_t = self.offset_ref_log_r_t_rv()
+        log_rtu_ref_subpop_in_week = log_r_t_in_weeks + jnp.where(
+            self.n_subpops > 1, offset_ref_log_r_t, 0
+        )
+
+        if self.n_subpops == 1:
+            i_first_obs_over_n_subpop = i_first_obs_over_n_ref_subpop
+            initial_exp_growth_rate_subpop = initial_exp_growth_rate_ref_subpop
+            log_rtu_subpop_in_week = log_rtu_ref_subpop_in_week[:, jnp.newaxis]
+        else:
```
damonbayer, line 237: For reviewing and onboarding people to the code, I think it would be helpful if there were a `.md` file or more comments throughout the code to explain what is going on and how data is structured.

E.g. if there is only 1 subpop, `i_first_obs_over_n_subpop` has shape x, but if there is more than one, it has shape y, where the dimension represent a and b`

damonbayer: For instance, I am curious why `log_rtu_subpop_in_week` requires `jnp.newaxis`, but `i_first_obs_over_n_subpop` and `initial_exp_growth_rate_subpop` do not.

sbidari:
> @sbidari Can you share the rendered qmd file somewhere for me to look at?

Rendered qmd files here https://cdc.sharepoint.com/:f:/r/teams/CenterforForecastingandOutbreakAnalytics/Shared%20Documents/General/02%20-%20Predict/Real%20Time%20Monitoring%20(RTM)%20Branch/Short%20Term%20Forecasts/PyRenew%20Wasewater%20Model?csf=1&web=1&e=3Y4JuM

demos/ww_model/model_comp.qmd
```diff
@@ -0,0 +1,157 @@
...
+  ggtitle("Vignette Data Model Comparison") +
+  scale_y_continuous("Hospital Admissions") +
+  scale_x_continuous("Time") +
+  theme(legend.position = "bottom")
+```
+
+## Rt Comparions
+
+```{r}
+combined_ci_for_plotting |>
```
damonbayer, line 102: These seems subtlety, but perhaps significantly different?

sbidari: They are definitely (noticeably) different. There is a difference in how the viral dynamics function is implemented in the stan model and the pyrenew model. Wanted to flag this in case you are not aware.

pyrenew_hew/ww_site_level_dynamics_model.py
sbidari: Sometimes I used `jnp.newaxis` to expand 1D array to ensure matrix operations work as intended.

In the part above, I extend `log_rtu_subpop_in_week` so that the code below works both when `n_subpops =1` and `n_subpops >1` rather than having separate codes for either case. But `initial_exp_growth_rate_subpop` works for either case without the need to add additional axis.
```
rtu_subpop = jnp.squeeze(
            jnp.repeat(
                jnp.exp(log_rtu_subpop_in_week),
                repeats=7,
                axis=0,
            )[:n_datapoints, :]
        )
```

demos/ww_model/model_comp.qmd
damonbayer: Difference in the mathematics? Or same math with different implementation? Is this documented somewhere?

pyrenew_hew/ww_site_level_dynamics_model.py
```diff
@@ -0,0 +1,538 @@
...
+        i0 = infection_initialization_process()
+        numpyro.deterministic("i0", i0)
+
+        inf_with_feedback_proc_sample = self.inf_with_feedback_proc.sample(
+            Rt=rtu_subpop,
+            I0=i0,
+            gen_int=generation_interval_pmf,
+        )
+
+        new_i_subpop = jnp.atleast_2d(
```
damonbayer, line 350: what does "new" indicate?

pyrenew_hew/ww_site_level_dynamics_model.py
```diff
@@ -0,0 +1,538 @@
...
+            jnp.concat(
+                [
+                    i0,
+                    inf_with_feedback_proc_sample.post_initialization_infections,
+                ]
+            )
+        )
+        if new_i_subpop.shape[0] == 1:
+            new_i_subpop = new_i_subpop.T
```
damonbayer, line 359: Can we do something different in the creation of `new_i_subpop` to avoid this?

pyrenew_hew/ww_site_level_dynamics_model.py
```diff
@@ -0,0 +1,538 @@
...
+
+        r_subpop_t = inf_with_feedback_proc_sample.rt
+        numpyro.deterministic("r_subpop_t", r_subpop_t)
+
+        state_inf_per_capita = jnp.sum(self.pop_fraction * new_i_subpop, axis=1)
```
damonbayer, line 364: would prefer a different word than `state` since we don't just run on states.

pyrenew_hew/ww_site_level_dynamics_model.py
```diff
@@ -0,0 +1,538 @@
...
+        numpyro.deterministic("latent_hospital_admissions", latent_hospital_admissions)
+
+        hospital_admission_obs_rv = NegativeBinomialObservation(
+            "observed_hospital_admissions", concentration_rv=self.phi_rv
+        )
+
+        if not is_predictive:
+            mu_obs_hosp = latent_hospital_admissions[self.hosp_times]
+        else:
+            mu_obs_hosp = latent_hospital_admissions
```
damonbayer, line 422: We should document why this is necessary

demos/ww_model/model_comp.qmd
sbidari: Different math. The stan model uses tent function approximation of the viral dynamics (https://github.com/CDCgov/ww-inference-model/blob/5caa395ab6d4f42f85a80469bbaca014e8b2878d/inst/stan/functions/observation_model.stan#L18), for the pyrenew model @dylanhmorris calculated actual values as difference of cdfs (https://github.com/CDCgov/pyrenew-hew/pull/49/files/44436bb0013723a4bc13599762567eeb42c50741#diff-87e0172f757eec7ac2a473bbda41b00e992c54cb83d16a5c17cabf4ee59b88b1)

pyrenew_hew/ww_site_level_dynamics_model.py
```diff
@@ -0,0 +1,538 @@
...
+            def batch_colvolve_fn(m):
+                return jnp.convolve(m, s, mode="valid")
+
+            model_net_i = jax.vmap(batch_colvolve_fn, in_axes=1, out_axes=1)(
+                new_i_subpop
+            )[-n_datapoints:, :]
+            numpyro.deterministic("model_net_i", model_net_i)
+
+            log10_g = self.log10_g_rv()
+            model_log_v_ot = (
```
damonbayer, line 444: This strikes me as a particularly confusing variable name. What are `v` and `ot` and why do we prefix with `model`?

pyrenew_hew/ww_site_level_dynamics_model.py
damonbayer, line 1: Overall, I wonder if it would be possible to modularize this model some more and perhaps add components to base pyrenew.

pyrenew_hew/ww_site_level_dynamics_model.py
sbidari: I can't think of alternative implementation that works with both 1D and 2D `new_i_subpop` but happy to implement if you have ideas.

damonbayer: Would `expand_dims` do this?

pyrenew_hew/ww_site_level_dynamics_model.py
```diff
@@ -0,0 +1,570 @@
...
+        return None
+
+    def validate(self):  # numpydoc ignore=GL08
+        return None
+
+    def sample(
+        self,
+        n_datapoints=None,
+        data_observed_hospital_admissions=None,
+        data_observed_log_conc=None,
```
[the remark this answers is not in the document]
damonbayer: `data_observed_log_conc`

damonbayer (approved):
Given that this runs and produces plausible results, I am going to merge to main now. I will open an issue for further improvements and evaluation to be done next year.

Issue #47, opened 2024-10-11
Posterior predictions for count proportions/percentages


SamuelBrand1:
## Goal

The goal of this issue is to add to existing observation processes to include proportion (or %) of some underlying total count are from a particular pathogen of interest.

## Context

CDC releases emergency dept visits dues to specific pathogen infection as %s of total visits, for example: [proportion of ED visits due to Covid](https://covid.cdc.gov/covid-data-tracker/#maps_percent-covid-ed). Therefore, it makes sense to be capable of observing, and nowcasting/predicting this target.

My first view on modelling this is that the lowest difficulty appropriate modelling approach is to have:

- A `pyrenew` model for count incidence of the target pathogen $E(t)$. Functionality for this exists already.
- An _independent_ model to predict total counts _less the target pathogen_ $N(t)$.

Then the proportion of interest is:

```math
\text{perc. total counts are target pathogen} = 100 \times \frac{E(t)}{E(t) + N(t)}.
```

Which we can sample from the posterior distribution of the model, either with an independent sample of $N(t)$ or a point est of $N(t)$.

## Alternative approaches to consider

Main alternative is to treat the proportion of interest as the target, rather than the approach above which treats the inference problem as a transformation on currently existing inference dependent on the sampling/point est of an independent model.

## Required features

- [ ] #53
- [ ] #69

## Specifications

Potentially to be added after investigation.

SamuelBrand1:
@damonbayer is #68 aimed at the completing the second task here?

PR #54, opened 2024-10-14
Issue 53: Model for prediction non-target pathogens

### LINKED ISSUE #53: A model to generate $N(t)$.
This issue is aimed at selecting a method for forecasting the _independent_ part of the denominator of the fraction:

```math
P(t) = {X(t) \over X(t) + N(t)}.
```

Where $X(t)$ is a target pathogen which is modelled using a sophisticated method e.g. `pyrenew-hew`, and $N(t)$ are the remainder of the numerator.

This approach assumes that independence of $X(t)$ is $N(t)$ is reasonable; for example when $N(t)$ is a mix of some other respiratory virus incidence but pre-dominantly non-infectious disease causes.

SamuelBrand1:
This PR closes #53 .

This PR adds two R functions in the folder `countpredr/R` along with associated unit tests. The added functionality:

- `generate_example_data`: This generates rough estimates of weekly ED visit counts based on publicly available data (see `README` and docstrings.
- `forecast_counts`: This generates forecasts of $N(t)$ using the `fable` package.

This PR also adds an example usage note rendered to `gfm` at `countpredr/README.md`.

## Forecast model

The forecast model is an ensemble of an ETS and a SARIMA model fitted on `log(count)` data (see `countpredr/README.md`). This is not a tuneable option, but rather an (IMO) robust forecaster for short term prediction of the non-target pathogen(s)/causes that we can use whilst focusing on forecasting the target pathogen.

Note that discrete degrees of freedom in model selection are auto-tuned over the default selection space by `fable::model`. I have changed the defaults for the ETS trend model so that a multiplicative option is part of the model selection space (`trend(method = c("N", "M", "A"))`).

## Context

This is part of resolving #47 by creating a forecast model for _non-target_ pathogens. The obvious next step is to add this to predictive modelling workflow, a simple example of doing this by joining to predictive samples is given in the README.


damonbayer (commented):
Thanks @SamuelBrand1! This looks really nice. Before I approve/merge I need to think a bit more about how we organize the various demos that are going into this project.

SamuelBrand1:
> Thanks @SamuelBrand1! This looks really nice. Before I approve/merge I need to think a bit more about how we organize the various demos that are going into this project.

I don't think you have an issue for this? 

I think this would be more effective with an issue laying out what your goals are, then we can add specifications to #47 .

damonbayer:
@SamuelBrand1 I think I will likely close this PR in favor of https://github.com/CDCgov/pyrenew-hew/pull/68

SamuelBrand1:
> @SamuelBrand1 I think I will likely close this PR in favor of #68

I think it might be worth merging this too. This could facilitate using the `fable` utilities as part of `scoringutilhelpers` (e.g. baseline forecaster).

PR #55, opened 2024-10-15
Modularize Post Processing

damonbayer:
This is a draft PR until my requested feature in `renv` is added. It is already in an open PR: https://github.com/rstudio/renv/pull/2018

The `renv` bit is just required for local development (and I could come up with a workaround if required). The modularized R script should already work in an azure context.

Closes https://github.com/CDCgov/pyrenew-hew/issues/40

PR #66, opened 2024-10-21
Make data prep and post-processing more generic

dylanhmorris:
Useful for cloud runs but I think good in general so worth merging now I think

### DIFF
--- a/nssp_demo/post_process.R
+++ b/nssp_demo/post_process.R
@@ -21,6 +21,9 @@ make_forecast_fig <- function(model_dir,
     pluck(1) %>%
     tail(1)
 
+  message(
+      glue::glue("Making forecast fig for {state_abb}/{disease_name_raw}"
+                 ))
 
   data_path <- path(model_dir, "data", ext = "csv")
   inference_data_path <- path(model_dir, "inference_data",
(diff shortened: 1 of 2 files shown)


nssp_demo/prep_data.py
```diff
@@ -181,20 +116,150 @@
...
-    with open(
-        pathlib.Path(state_folder, "data_for_model_fit.json"), "w"
-    ) as json_file:
+    with open(os.path.join(state_dir, "data_for_model_fit.json"), "w"
+              ) as json_file:
         json.dump(data_for_model_fit, json_file)
 
-logger.info("Data preparation complete.")
+
+def main(disease,
```
damonbayer, line 133: We can either get rid of `main` and warp its contents in  `if __name__ == "__main__": ` or just wrap the call to `main`.

nssp_demo/post_process.R
damonbayer, line 1: I think this is basically redundant with https://github.com/CDCgov/pyrenew-hew/pull/55

dylanhmorris: agree. I was not sufficiently aware of #55 and also needed to fix some things.

dylanhmorris: I agree. Can revert once #55 lands.

PR #68, opened 2024-10-21
Denominator Forecasting

damonbayer:


SamuelBrand1:
This PR will also close #69 .

nssp_demo/total_ed_admissions_forecast.R
dylanhmorris, line 1: This should be total_minus_target, no?

damonbayer: I think you're right. This should actually model `total - target`, and then the percent is `target / (total - target + target)`

nssp_demo/total_ed_admissions_forecast.R
```diff
@@ -0,0 +1,84 @@
...
+    select(date, .draw, total_ED_admissions = ED_admissions)
+
+  forecast_samples
+}
+
+main <- function(model_dir, n_forecast_days = 28, n_samples = 2000) {
+  # to do: do this with json data that has dates
+  data_path <- path(model_dir, "data", ext = "csv")
+
+  denom_data <- read_csv(data_path) %>%
```
damonbayer, line 59: this name is not accurate

dylanhmorris (approved):
LGTM, thanks @damonbayer!

PR #78, opened 2024-10-28
Create single location end-to-end forecasting CLI and associated container image

dylanhmorris:
This PR:

- Creates a script (`forecast_state.py`) that runs a forecast end to end (from data prep through post-processing) and writes the results (including intermediate products) to designated locations.
- Creates an OCI container to allow that script to be run easily on cloud machines (e.g. Azure batch nodes).
- Adds scripts under `nssp_demo/batch` to set up pools and jobs.
- Transitions from using `renv` to manage R deps to creating a skeleton package (`hewr`) with all needed R deps as `Imports`, and installs it in as part of the container build.
- Does some cleanup on scripts
- Fixes terminology around "ed visits" for clarity / precision.
- Reduces default infection feedback priors.
- All of the above has been tested in Azure batch.


nssp_demo/fit_model.py
```diff
@@ -3,43 +3,57 @@
...
+
+    with open(
+        model_dir / "posterior_samples.pickle",
+        "wb",
+    ) as file:
+        pickle.dump(my_model.mcmc, file)
+
+
+if __name__ == "__main__":
+    n_chains = 4
```
damonbayer, line 45: Should all arguments to `fit_and_save_model` be read from the command line, with defaults?

nssp_demo/forecast_non_target_visits.R
```diff
@@ -0,0 +1,116 @@
...
+  "stringr",
+  "fs",
+  "fable",
+  "jsonlite",
+  "argparser",
+  "arrow"
+)
+
+## load in packages without messages
+for (package in script_packages) {
```
damonbayer, line 15: I prefer `purrr::walk` for compactness or `sapply` if you want to stay in base.

nssp_demo/forecast_non_target_visits.R
```diff
@@ -0,0 +1,116 @@
...
+  add_argument(
+    "--n-samples",
+    help = "Number of samples to generate",
+    default = 2000L
+  )
+
+argv <- parse_args(p)
+model_dir <- path(argv$model_dir)
+n_forecast_days <- argv$n_forecast_days
+n_samples <- argv$n_samples
```
damonbayer, line 101: I just learned about [`list2env`](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/list2env), which does this.

nssp_demo/forecast_state.py
```diff
@@ -0,0 +1,222 @@
...
+import polars as pl
+from prep_data import process_and_save_state
+
+numpyro.set_host_device_count(4)
+
+from fit_model import fit_and_save_model  # noqa
+from generate_predictive import generate_and_save_predictions  # noqa
+
+
+def forecast_denominator(model_dir: Path, n_forecast_days: int) -> None:
```
damonbayer, line 18: Since this is run in the same place as the numerator forecasting, we should revise to guarantee that they produce the same number of samples.

nssp_demo/forecast_state.py
```diff
@@ -0,0 +1,222 @@
...
+import os
+import subprocess
+from datetime import datetime, timedelta
+from pathlib import Path
+
+import numpyro
+import polars as pl
+from prep_data import process_and_save_state
+
+numpyro.set_host_device_count(4)
```
damonbayer, line 12: Consider reading this from the command line or setting to something very high by default.

nssp_demo/forecast_state.py
damonbayer: Since we're mixing languages, I wonder if you could just do an inline Rscript to detect the devices with [`parallelly::availableCores`](https://parallelly.futureverse.org/reference/availableCores.html).

nssp_demo/forecast_state.py
damonbayer, line 1: It would be nice to be able to skip steps (or start at a certain step) based on some command line arg. Then, for example, you could trivially rerun the post-processing without refitting the model.

damonbayer, line 1: Should we have a more descriptive directory name to indicate something about when the model was run or what version of the code was used? or just a unique identifier?

nssp_demo/forecast_state.py
```diff
@@ -0,0 +1,222 @@
...
+    if report_date == "latest":
+        report_date = max(
+            f.stem for f in Path(nssp_data_dir).glob("*.parquet")
+        )
+
+    report_date = datetime.strptime(report_date, "%Y-%m-%d").date()
+
+    logger.info(f"Report date: {report_date}")
+
+    last_training_date = report_date - timedelta(days=1)
```
damonbayer, line 69: Should `last_training_date` be configurable from an argument (with default 1)?

nssp_demo/forecast_state.py
```diff
@@ -0,0 +1,222 @@
...
+
+    datafile = f"{report_date}.parquet"
+    nssp_data = pl.scan_parquet(Path(nssp_data_dir, datafile))
+    param_estimates = pl.scan_parquet(Path(param_data_dir, "prod.parquet"))
+    model_dir_name = (
+        f"{disease.lower()}_r_{report_date}_f_"
+        f"{first_training_date}_t_{last_training_date}"
+    )
+
+    model_data_dir = Path(output_data_dir, model_dir_name)
```
damonbayer, line 83: Throughout, I have generally used `model_dir` to refer to a particular state folder and `base_dir` to refer to the folder that contains the state folders (each being a `model_dir`). We can change these names (as I don't think they're good), but we should be consistent.

nssp_demo/generate_predictive.py
```diff
@@ -3,72 +3,70 @@
...
+
+    idata.to_dataframe().to_csv(model_dir / "inference_data.csv", index=False)
+
+    # Save one netcdf for reloading
+    idata.to_netcdf(model_dir / "inference_data.nc")
+
+    # R cannot read netcdf files with groups,
+    # so we split them into separate files.
+    for group in idata._groups_all:
+        idata[group].to_netcdf(model_dir / f"inference_data_{group}.nc")
```
damonbayer, line 49: This can be removed. It is no longer relevant, based on our plans for using forecasttools to convert idata to tidy_draws.

nssp_demo/postprocess_state_forecast.R
```diff
@@ -156,86 +137,101 @@ make_one_forecast_fig <- function(target_disease,
...
+      data_type = col_character(),
+      ed_visits = col_double(),
+      date = col_date()
+    )
+  ) |>
+    mutate(
+      disease = if_else(
+        disease == disease_name_nssp,
+        "Disease", # assign a common name for
+        ## use in plotting functions
```
damonbayer, line 168:
```suggestion
        # use in plotting functions
```

nssp_demo/postprocess_state_forecast.R
```diff
@@ -251,9 +247,50 @@ make_forecast_figs <- function(model_dir,
...
+  add_argument(
+    "--good-chain-tol",
+    help = "Tolerance level for determining good chains",
+    default = 2L
+  )
+
+argv <- parse_args(p)
+model_dir <- path(argv$model_dir)
+filter_bad_chains <- argv$filter_bad_chains
+good_chain_tol <- argv$good_chain_tol
```
damonbayer, line 282: another opportunity for [`list2env`](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/list2env)

nssp_demo/postprocess_state_forecast.R
```diff
@@ -265,47 +302,3 @@ iwalk(forecast_figs, ~ save_plot(
...
-  )
-
-  iwalk(forecast_figs, ~ save_plot(
-    filename = path(model_dir, glue("{.y}_forecast_plot"), ext = "pdf"),
-    plot = .x,
-    device = cairo_pdf, base_height = 6
-  ))
-})
-
-# Combine figures across states
```
damonbayer, line 294: This functionality should probably be recreated somewhere in a bash script.

damonbayer (approved):
I left many comments, but no mandatory changes. Let me know if you would like me to address any of them myself.

Makefile
```diff
@@ -0,0 +1,17 @@
+.PHONY: container_build container_tag acr_login container_push
+
+ENGINE := docker
+CONTAINER_NAME := pyrenew-hew
+CONTAINER_REMOTE_NAME := $(ACR_TAG_PREFIX)$(CONTAINER_NAME)":latest"
```
gvegayon, line 5: I would add this as optional perhaps? See for instance: https://github.com/CDCgov/cfa-epinow2-pipeline/blob/338f0ec947084f5201b68d1be7c55bd0169e07d0/Makefile#L2-L4

nssp_demo/batch/setup_job.py
```diff
@@ -0,0 +1,88 @@
...
+        "--nssp-data-dir nssp_demo/nssp-etl/gold "
+        "--param-data-dir nssp_demo/params "
+        "--output-data-dir nssp_demo/private_data"
+        "'"
+    )
+
+    states = pl.read_csv(
+        "https://raw.githubusercontent.com/k5cents/usa/"
+        "refs/heads/master/data-raw/states.csv"
+    )
```
gvegayon, line 56: I would replace this for a hashed version (permanent link) to that content. Ideally, I would also point to a file within pyrenew-hew.

nssp_demo/forecast_state.py
gvegayon: But this is a demo, so I would just pick 2 and leave that as a default with `npus = 2 # set this accordingly` or whatnot.

Containerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3.12
+
+RUN apt-get update
+RUN apt-get install -y r-base
```
gvegayon, line 4: This could be somewhat risky as you are picking whatever is the current version of r-base supported by apt-get. Why not point to a fixed version instead? Also, why not using one of the rocker images instead?

pyproject.toml
```diff
@@ -15,6 +15,7 @@ jupyter = "^1.0.0"
 ipykernel = "^6.29.5"
 polars = "^1.5.0"
 pyarrow = "^17.0.0"
+azuretools = {git = "https://github.com/cdcent/cfa-stf-azuretools"}
```
damonbayer, line 18: We can make this depend on some secret https://github.com/orgs/python-poetry/discussions/3794

nssp_demo/fit_model.py
dylanhmorris: addressed in 28a529d

nssp_demo/forecast_non_target_visits.R
dylanhmorris: done in e01bfd9

nssp_demo/forecast_non_target_visits.R
dylanhmorris: Maybe I'm not understanding what you're suggesting, but I don't think the readability /  safety versus concision tradeoff is in its favor with only three args

nssp_demo/forecast_state.py
dylanhmorris: Out of scope for this PR but a nice idea. Question in my mind about is whether it's worth the lift. Depends how certainly and soon we plan to move away from manual workflow specification (toward tools like airflow)

dylanhmorris: Can discuss in #81

nssp_demo/forecast_non_target_visits.R
damonbayer: That's a reasonable conclusion.

nssp_demo/generate_predictive.py
dylanhmorris: Done in 9825c2d

nssp_demo/batch/setup_job.py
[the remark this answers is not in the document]
dylanhmorris: Definitely a TODO

nssp_demo/batch/setup_job.py
dylanhmorris: Long term I think we should point at https://www2.census.gov/geo/docs/reference/state.txt or a vintage thereof.

damonbayer: That does not included population size, which is what we need.

nssp_demo/forecast_state.py
dylanhmorris: I think we should standardize on `model_run_dir` as variable name and `model-run-dir` / `--model-run-dir` as the associated command line argument.

nssp_demo/forecast_state.py
dylanhmorris: Eventually yes, out of scope for now.

nssp_demo/forecast_state.py
dylanhmorris: In general should figure out the best way to coordinate parallelization settings.

nssp_demo/batch/setup_job.py
dylanhmorris: Oops, this one https://www2.census.gov/geo/docs/reference/cenpop2020/CenPop2020_Mean_ST.txt

nssp_demo/forecast_state.py
dylanhmorris: To discuss in #85

Makefile
dylanhmorris: Azure CR expects the engine to be named docker (I'm actually using an aliased podman here)

nssp_demo/forecast_state.py
damonbayer: Somehow this thread got way of track from my original comment, which, I believe, remains unaddressed.

dylanhmorris: Original request addressed in 8062222

dylanhmorris:
Merging this @damonbayer. Thanks for the review / comments.

Issue #57, opened 2024-10-16
Reorganize repository based on current plans


damonbayer:
There are a few issues with the current repo structure:

1. Our NSSP model started as a "demo," hence the `nssp_demo` folder, but it is now the main offering of the repo.
2. We are collecting a variety of other "demo" scripts, currently stored in the `notebooks` folder.

I would like to reorganize the repo to make things a bit more clear. I think I will have a better idea of how to do that once we have run things in azure.

SamuelBrand1:
My 2p worth, and this might be better implemented over more than one repo:

Have a structure of:

- `tools`: thats `pyrenew-hew` etc which do the analysis. This should be callable from projects in this repo.
- `projects`: thats discrete projects we want to work on in the open e.g. `nssp`.
- `demos` -> `docs` as vignettes.

damonbayer:
Current thinking:

- `pyrenew-hew`: setup as a python library that has the PyRenew model(s)
- `utils_R`: R utility scripts that can be `source`d or could be set up as an R package
- `utils_py`:
- `pipeline`: scripts related to running the pipeline
  - could be further separated into different `pipeline` folders if we have different pipelines
- `demos`: vignettes / docs

dylanhmorris:
Largely agree, but I think we can and should avoid `import`-able/`source`-able middle ground between pipeline scripts and packaged code. So I'd suggest the above except:

- Bundle anything reused enough to be a `utils-py` into `pyrenew-hew.utils`
- Any `utils_R` that don't belong in `pipeline` could be a micro-package called `hewR` or something

SamuelBrand1:
- [ ] `pyrenew-hew`: setup as a python library that has the PyRenew model(s)
- [ ] `utils_R`: R utility scripts that can be `source`d or could be set up as an R package
- [ ] `utils_py`: Python utility scripts that can be `import`ed.
- [ ] `pipeline`: scripts related to running the pipeline
  - could be further separated into different `pipeline` folders if we have different pipelines
- [ ] `demos`: vignettes / docs

PR #87, opened 2024-10-30
Infrastructure for running on vintaged NSSP data

dylanhmorris:
Including the ability to mix facility-level recent vintages and earlier aggregate vintages, with checks to ensure data is deduplicated and also complete.


nssp_demo/batch/setup_job.py
```diff
@@ -63,14 +72,27 @@ def main(job_id, pool_id, container_image) -> None:
...
-            task = get_task_config(
-                f"{job_id}-{state}-{disease}",
-                base_call=base_call.format(state=state, disease=disease),
-                container_settings=container_settings,
-            )
-            client.task.add(job_id, task)
+    report_dates = [
+        datetime.date(2023, 10, 11) + datetime.timedelta(weeks=x)
+        for x in range(30)
+    ]
```
damonbayer, line 78: I would prefer a polars date_range so we can more easily discern/configure the end date.

damonbayer:
We should really think about file organization here or immediately after merging. Perhaps, put some of this in a "projects" folder, as suggested here https://github.com/CDCgov/pyrenew-hew/issues/57#issuecomment-2417437301

nssp_demo/batch/setup_job.py
```diff
@@ -63,14 +72,27 @@ def main(job_id, pool_id, container_image) -> None:
...
+
+    for disease in ["Influenza"]:
```
damonbayer, line 80: When we do COVID retro, how do you imagine that being implemented? A different `setup_job`?

nssp_demo/batch/setup_pool.py
```diff
@@ -36,6 +36,7 @@ def main(pool_name: str) -> None:
     pool_config.mount_configuration = blob.get_node_mount_config(
         storage_containers=[
             "nssp-etl",
```
damonbayer, line 38: Is nssp-etl used at all here?

nssp_demo/forecast_non_target_visits.R
```diff
@@ -99,7 +99,7 @@ n_forecast_days <- argv$n_forecast_days
 n_samples <- argv$n_samples
 
 disease_name_nssp_map <- c(
-  "covid-19" = "COVID-19/Omicron",
+  "covid-19" = "COVID-19",
```
damonbayer, line 102: I thought we had agreed that our new vintages would use `COVID-19/Omicron` to be consistent with the modern vintages?

nssp_demo/forecast_state.py
```diff
@@ -49,26 +49,41 @@ def main(
...
+    state_level_nssp_data_dir: Path | str,
     param_data_dir: Path | str,
     output_data_dir: Path | str,
     n_training_days: int,
     n_forecast_days: int,
     n_chains: int,
     n_warmup: int,
     n_samples: int,
-    last_training_date: str,
+    last_training_date: str = None,
```
damonbayer, line 61: I think I would prefer to parameterize this in terms of days prior to `report_date`

nssp_demo/forecast_state.py
```diff
@@ -49,26 +49,41 @@ def main(
...
 ):
     logging.basicConfig(level=logging.INFO)
     logger = logging.getLogger(__name__)
 
+    available_facility_level_reports = [
+        datetime.strptime(f.stem, "%Y-%m-%d").date()
+        for f in Path(facility_level_nssp_data_dir).glob("*.parquet")
```
damonbayer, line 68: Could use a more specific glob pattern to ensure ` datetime.strptime` will not fail by being called on non-compliant files (though that shouldn't happen).

nssp_demo/forecast_state.py
```diff
@@ -49,26 +49,41 @@ def main(
...
-        )
-    report_date = datetime.strptime(report_date, "%Y-%m-%d").date()
+        report_date = max(available_facility_level_reports)
+    else:
+        report_date = datetime.strptime(report_date, "%Y-%m-%d").date()
+
+    if report_date in available_state_level_reports:
+        state_report_date = report_date
+    else:
+        state_report_date = max(available_state_level_reports)
```
damonbayer, line 83: Why?

nssp_demo/forecast_state.py
```diff
@@ -87,12 +102,34 @@ def main(
...
+
+    if report_date in available_facility_level_reports:
+        logger.info(
+            "Facility level data available for " "the given report date"
+        )
+        facility_datafile = f"{report_date}.parquet"
+        facility_level_nssp_data = pl.scan_parquet(
+            Path(facility_level_nssp_data_dir, facility_datafile)
+        )
+    if state_report_date in available_state_level_reports:
```
damonbayer, line 122: This is always true because of https://github.com/CDCgov/pyrenew-hew/pull/87/files#r1823063547. Is that intended?

nssp_demo/prep_data.py
```diff
@@ -8,43 +8,126 @@
...
         .group_by(["reference_date", "disease"])
         .agg(pl.col("value").sum().alias("ed_visits"))
         .with_columns(
-            pl.when(pl.col("reference_date") <= last_training_date)
-            .then(pl.lit("train"))
-            .otherwise(pl.lit("test"))
-            .alias("data_type")
+            disease=pl.col("disease")
+            .cast(pl.Utf8)
+            .replace({v: k for k, v in disease_map.items()}),
```
damonbayer, line 93: I think you can just use `.replace(disease_map)`

From [polars docs](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.replace.html):

```python
mapping = {2: 100, 3: 200}
df.with_columns(replaced=pl.col("a").replace(mapping))
```

nssp_demo/prep_data.py
```diff
@@ -94,9 +177,6 @@ def process_and_save_state(
             & (pl.col("disease") == disease)
             & (pl.col("parameter") == "right_truncation")
             & (pl.col("end_date").is_null())
-            & (
-                pl.col("reference_date") <= report_date
-            )  # estimates nearest the report date
```
damonbayer, line 99: Why is this removed?

nssp_demo/forecast_non_target_visits.R
dylanhmorris: Vintages yes, but in the version of `prep_data.py` in this PR I change `COVID-19/Omicron` to `COVID-19` so we can use `COVID-19` throughout downstream.

nssp_demo/forecast_state.py
dylanhmorris: I went back and forth on this. Happy to switch to measuring it relative to the report date.

nssp_demo/priors.py
```diff
@@ -25,7 +25,7 @@
 )
 
 eta_sd_rv = DistributionalVariable(
-    "eta_sd", dist.TruncatedNormal(0, 0.01, low=0)
+    "eta_sd", dist.TruncatedNormal(0.04, 0.02, low=0)
```
damonbayer, line 28: Why?

nssp_demo/prep_data.py
dylanhmorris: I am inverting the `disease_map` here (note `{v: k for v, v}`). Perhaps for readability I should define the inverse right along with the mapn and then call this with `replace(inverse_disease_map)`

nssp_demo/priors.py
dylanhmorris: https://github.com/CDCgov/wastewater-informed-covid-forecasting/pull/198

nssp_demo/prep_data.py
dylanhmorris: For report dates far enough in the past, there will not be a PMF with an appropriate reference date, and since we were planning not to attempt right-truncuation modeling for the most recent data, I thought simplest might be to use a fixed PMF. But we could of course always use the earliest PMF for `report_dates` before the first PMF `reference_date` (or fit PMFs for the earlier period based on the vintages! but we shouldn't try to do that IMO)

nssp_demo/forecast_state.py
dylanhmorris, line 83: To use the latest state data when hybridizing, but I see now that a more correct thing would be 

```suggestion
    elif report_date > max(available_state_level_reports):
        state_report_date = max(available_state_level_reports)
     else:
        state_report_date = None
```
And then error if neither the facility-level nor the state level has a sufficiently early recorded `report_date`

nssp_demo/prep_data.py
damonbayer: Please document as an issue.

dylanhmorris: #88

dylanhmorris:
> We should really think about file organization here or immediately after merging. Perhaps, put some of this in a "projects" folder, as suggested here https://github.com/CDCgov/pyrenew-hew/issues/57#issuecomment-2417437301

Agree, I think this is fairly high priority once this lands.

nssp_demo/prep_data.py
dylanhmorris: Done in 7bb74d7

nssp_demo/forecast_state.py
dylanhmorris: I think we can leave this for now, but in fddcbfe I have created a `get_available_reports` helper function that we could modify in the future to have different (potentially stricter) logic.

nssp_demo/forecast_state.py
dylanhmorris: Changed in 9e87af9

nssp_demo/batch/setup_pool.py
dylanhmorris: Yes, this tells Batch that when nodes are spun up in this pool they should mount this blob storage container, which is useful because it contains the (later) vintages.

nssp_demo/batch/setup_job.py
dylanhmorris: In this case I think a number of weeks is fairly useful to reason about, but we can change going forward.

nssp_demo/forecast_state.py
dylanhmorris: Improved the logic so that it's no longer true (in particular if you try to grab data from _before_ the state level report dates start.

dylanhmorris: Longer run we should perhaps refactor and generalize this data hybridization and date picking structure, but I'm reluctant to spend that time now given that other things may change.

nssp_demo/forecast_state.py
dylanhmorris: This should be clearer (and more correct) with the improved logic in 2016138

We wish to fall back on the latest state-level data when the report date is after the last available state level report

nssp_demo/prep_data.py
dylanhmorris: We should still consider, but for now a solution with date math in 0543961

dylanhmorris:
Ready for re-review @damonbayer

nssp_demo/forecast_state.py
```diff
@@ -45,38 +45,68 @@ def postprocess_forecast(model_run_dir: Path) -> None:
...
+    elif report_date > first_available_state_report:
+        raise ValueError(
+            "Dataset appear to be missing some state-level "
+            f"reports. First entry is {first_available_state_report}, "
+            f"last is {last_available_state_report}, but no entry "
+            f"for {report_date}"
         )
-    report_date = datetime.strptime(report_date, "%Y-%m-%d").date()
+    else:
+        state_report_date = None
```
damonbayer, line 102: This is only reached when `report_date < first_available_state_report`, so why not error instead of None?

dylanhmorris: because in the future we might have facility date that's more out of date than state data (admittedly elsewhere in the code we tacitly assume the facility-level data is more recent.

damonbayer: I think I would rather explicitly assume that the facility-level data is more recent so that things aren't silently wrong if that changes.

dylanhmorris: Works for me.

nssp_demo/prep_data.py
```diff
@@ -183,29 +298,66 @@ def main(
...
+        )
+        facility_datafile = f"{report_date}.parquet"
+        facility_datapath = Path(
+            facility_level_nssp_data_dir, facility_datafile
+        )
+        with open(facility_datapath) as facility_file:
+            facility_level_nssp_data = pl.scan_parquet(facility_file)
+        dat = facility_level_nssp_data
+    if state_report_date in available_state_level_reports:
+        logger.info("State-level data available for the " "given report date.")
```
damonbayer, line 343:
```suggestion
        logger.info("State-level data available for the given report date.")
```

nssp_demo/prep_data.py
damonbayer, line 1: Should we just get rid of `main` here?

damonbayer: or otherwise try to DRY `prep_data.main` and `forecast_state.main`?

damonbayer (approved):
One provocative suggestion, but I'm okay with merging.

nssp_demo/prep_data.py
dylanhmorris: Good point.

dylanhmorris: done in f49a962

nssp_demo/forecast_state.py
dylanhmorris: Done in e9abdd6

PR #93, opened 2024-10-31
Issue 92: add baseline timeseries forecaster for target

SamuelBrand1:
This PR adds a baseline forecaster for the numerator of the ed visit target pathogen.

## Modifications

- The `fit_and_forecast` function now accepts a choice of col name to run the timeseries forecaster (ensembled ETS + SARIMA) on. It also accepts an output name.
- `fit_and_forecast` function now only fits this model, instead of the ensemble model _and_ the base models separately. This was a legacy of visual comparison script that is not needed.
- The `main` function script now produces both a denominator and numerator forecast. To reflect this I've renamed to `timeseries_forecasts.R`. The assumption is that we'll always want this baseline given we are using the same method for numerator.
- The `forecast_state.py` script has been updated to point at the new script name.

## Not modified

This PR does not include post-processing of the timeseries forecaster. The numerator forecast is unchanged so this is not necessary and will be done in a follow up PR.

## Potential pain points

I've inferred the data schema for this PR, and this PR does not commit a unit test for the modified function.


nssp_demo/forecast_state.py
```diff
@@ -189,7 +189,7 @@ def main(
 
     logger.info("Performing non-target pathogen forecasting...")
```
dylanhmorris, line 190:
```suggestion
    logger.info("Performing baseline forecasting and non-target pathogen forecasting...")
```

dylanhmorris (approved):
LGTM. We just need to tweak one info message, which I will do. Thanks @SamuelBrand1!

PR #97, opened 2024-10-31
Issue 94: CDC flat forecaster for ED visits

### LINKED ISSUE #94: CDC flat baseline
This is a task of #80 .

The goal of this task is to use the baseline forecaster https://cmu-delphi.github.io/epipredict/reference/cdc_baseline_forecaster.html to produce a useful comparison baseline of the _proportion_ of a target pathogen.

Because this is a simple flatline, my view is that this is easiest achieved with a logit-transform on the proportion data, forecasting that and then gathering the prediction quantiles.

## Alternative
We can follow the same pattern as the main forecasting approach: that is forecasting the numerator and using an existing denominator forecast to construct proportions.

SamuelBrand1:
This PR adds the CDC flat forecaster to the timeseries forecast script. This PR closes #94.

## Approach
After some consideration I've gone for the _alternative_ approach described in #94. That is the cdc flat forecaster is forecasting numbers of ED visits for the target pathogen.

This was not my initial favoured approach, however, the data frequency is daily and the natural forecast output was in quantiles. One approach could be to weekly bin the numerator and denominator, then generate weekly proportions and then forecast using flat forecaster, but I think this method is more straightforward.

## Other additions

- `epipredict` and `epiprocess` need to be installed with `pak`. I've added `pak` and these to the `hewr` deps and added a conditional install step in the script. If this is unnecessary or a smoother option exists please point this out.
- Coercion to `as_tsibble` has been moved into `fit_and_forecast` because the cdc flat forecaster does a different coercion to `epi_df`.

## Not modified

This PR does not include post-processing of the cdc flat forecaster. The numerator forecast is unchanged so this is not necessary and will be done in a follow up PR.

## Potential pain points

I've inferred the data schema for this PR, and this PR does not commit a unit test for the modified function.

### DIFF
--- a/hewr/DESCRIPTION
+++ b/hewr/DESCRIPTION
@@ -29,4 +29,7 @@ Imports:
     tibble,
     tidybayes,
     tidyr,
-    urca
+    urca,
+    pak,
+    epipredict,
+    epiprocess
--- a/nssp_demo/timeseries_forecasts.R
+++ b/nssp_demo/timeseries_forecasts.R
@@ -8,14 +8,36 @@ script_packages <- c(
   "fable",
   "jsonlite",
   "argparser",
-  "arrow"
+  "arrow",
+  "pak",
+  "glue"
 )
+
+script_pak_packages <- c(
+  "epipredict",
+  "epiprocess"
+)
+##
 
 ## load in packages without messages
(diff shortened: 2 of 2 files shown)


dylanhmorris:
Closing this as it now forms part of #102

PR #115, opened 2024-11-12
Fix: fix arg name difference 

SamuelBrand1:
This small change fixes arg name difference between `fit_model.py` and `loop_fit.sh` ~in favour of the shell script~ in favour of py script (fewer changes).


damonbayer (approved):

dylanhmorris:
Thanks, @SamuelBrand1!

PR #129, opened 2024-11-15
Issue 127: add offset to ts models

### LINKED ISSUE #127: Investigate NA forecasts
`pyrenew-test-output/influenza_r_2024-11-12_f_2024-08-29_t_2024-11-11/AK/baseline_ts_count_ed_visits_forecast.parquet`

forecast column appears to be entirely NA

SamuelBrand1:
This PR closes #127.

@damonbayer  identified that the problem with `NA` forecast output was occurring in numerator forecasts when the data hit zero due to `log` transform.

I've added an offset following identical pattern (1 / max observed to symmetrise in log domain) to elsewhere in this repo.

### DIFF
--- a/pipelines/timeseries_forecasts.R
+++ b/pipelines/timeseries_forecasts.R
@@ -71,14 +71,20 @@ fit_and_forecast <- function(data,
   forecast_horizon <- glue::glue("{n_forecast_days} days")
   target_sym <- rlang::sym(target_col)
   output_sym <- rlang::sym(output_col)
+
+  max_visits <- data |>
+    pull(!!target_sym) |>
+    max(na.rm = TRUE)
+  offset <- 1 / max_visits
+
   fit <-
     data |>
     as_tsibble(index = date) |>
     filter(data_type == "train") |>
     model(
       comb_model = combination_ensemble(
-        ETS(log(!!target_sym) ~ trend(method = c("N", "M", "A"))),
-        ARIMA(log(!!target_sym))
+        ETS(log(!!target_sym + offset) ~ trend(method = c("N", "M", "A"))),
+        ARIMA(log(!!target_sym + offset))
       )
     )
 


dylanhmorris:
I would've thought offset of 1 for raw counts; we're using 1 / max_visits as the offset for proportions

damonbayer:
@dylanhmorris Agree that it's not consistent with what we do in the scoring, but an offset of 1 could be relatively big since the case counts can be so low.

dylanhmorris:
> @dylanhmorris Agree that it's not consistent with what we do in the scoring, but an offset of 1 could be relatively big since the case counts can be so low.

Just subtract 1 when you convert the forecast back to natural scale? (i.e. we explicitly forecast the quantity log(1+x) and then compute x)

damonbayer:
@dylanhmorris Yes. Lets actually "undo" the offset here after fitting the model, before saving the results.

Somewhere in here:
https://github.com/CDCgov/pyrenew-hew/blob/dbcfb6b862530feb454613934e359c1915678a8a/pipelines/timeseries_forecasts.R#L85-L89

dylanhmorris:
@SamuelBrand1 can you implement?

SamuelBrand1:
> I would've thought offset of 1 for raw counts; we're using 1 / max_visits as the offset for proportions

The same underlying idea applies I think: in `log` domain the maximum value is `log(max_visits)` and the minimum is `-log(max_visits)` if we have a zero in the natural domain. I assumed this was the idea, its symmetrises the log-domain data in the presence of zeros which seems a good idea irrespective of whether its counts or proportions?

SamuelBrand1:
> > @dylanhmorris Agree that it's not consistent with what we do in the scoring, but an offset of 1 could be relatively big since the case counts can be so low.
> 
> Just subtract 1 when you convert the forecast back to natural scale? (i.e. we explicitly forecast the quantity log(1+x) and then compute x)

`fable` does the backtransform automatically in the case of `log(x + offset)` transforms https://fable.tidyverts.org/articles/transformations.html . Although, for more complex transforms you have to wrap the transform with its backtransform in [`fabletools::new_transformation`](https://fabletools.tidyverts.org/reference/new_transformation.html).

SamuelBrand1:
> @SamuelBrand1 can you implement?

As above, I think the back-transform is done automatically... Do you want this for the pyrenew output?

damonbayer (approved):
@SamuelBrand1 fable is very impressive! Thanks!

PR #130, opened 2024-11-15
Separate Tidying MCMC from generating figures

damonbayer:


dylanhmorris (approved):
Have confirmed this works as expected with an individual state and with the national timeseries.

PR #117, opened 2024-11-12
issue 116: Generate test posterior samples

SamuelBrand1:
This PR closes #116.

This is a simple PR that tidies up the local fit loop utility, and creates a test directory which can be aimed at with `loop_fit.sh`. `.pickle` files are ignored so a test run with unit tests should start with this step.

## Run fake posterior generation

```bash
% bash ./iteration_helpers/loop_fit.sh ./tests/run_test_inference
```

from `/pipelines`.

This will be part of planned test runs.

PR #131, opened 2024-11-18
Update to run posterior predictive

SamuelBrand1:
This small PR merges a small change to #117 so that _both_ the posterior distribution from the pyrenew model can be generated _and_ the posterior predictive quantities.

### DIFF
--- a/pipelines/generate_predictive.py
+++ b/pipelines/generate_predictive.py
@@ -53,7 +53,7 @@ if __name__ == "__main__":
         description=("Do posterior prediction from a pyrenew-hew fit.")
     )
     parser.add_argument(
-        "--model_run_dir",
+        "model_run_dir",
         type=Path,
         help=(
             "Path to a directory containing the model fitting data "
--- a/pipelines/generate_predictive.py
+++ b/pipelines/generate_predictive.py
@@ -53,7 +53,7 @@ if __name__ == "__main__":
         description=("Do posterior prediction from a pyrenew-hew fit.")
     )
     parser.add_argument(
-        "model-run-dir",
+        "--model_run_dir",
         type=Path,
         help=(
             "Path to a directory containing the model fitting data "
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -1,6 +1,8 @@
 name: Create Docker Image
 
 on:
+  push:
+    branches: [main]
   pull_request:
   workflow_dispatch:
 
--- a/.gitignore
+++ b/.gitignore
@@ -394,6 +394,6 @@ poetry.lock
 notebooks/*_files/
 notebooks/*.md
 
-nssp_demo/private_data/*
+private_data/*
 *_files/
 .vscode/settings.json
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 *.xlsx
 *.rds
 *.pickle
+*.nc
 
 # Documents
 *.doc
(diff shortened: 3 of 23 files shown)


damonbayer:
@SamuelBrand1 Can you clarify the reason for these changes?

I'm also seeing the warning about the container file again. I'm wondering if that is because `add-test-data` isn't up to date with `main`.

SamuelBrand1:
> @SamuelBrand1 Can you clarify the reason for these changes?
> 
> I'm also seeing the warning about the container file again. I'm wondering if that is because `add-test-data` isn't up to date with `main`.

I'm basing the test output generation of the `bash` scripts inside of `iteration_helpers` which are not required for the `azure` run but call down onto the same functions. This PR merges into `add-demo-data` branch a small fix that gets the posterior predictive functions to run from the bash script.

pipelines/generate_predictive.py
```diff
@@ -53,7 +53,7 @@ def generate_and_save_predictions(
         description=("Do posterior prediction from a pyrenew-hew fit.")
     )
     parser.add_argument(
-        "model-run-dir",
+        "--model_run_dir",
```
damonbayer, line 56:
```suggestion
        "model_run_dir",
```

I think this would be my preference. Maybe we could sneak in a change throughout the repo:

- [Instances of `"model-run-dir"`](https://github.com/search?q=repo%3ACDCgov%2Fpyrenew-hew%20model-run-dir&type=code)
- [Instances of `"--model_run_dir"`](https://github.com/search?q=repo%3ACDCgov%2Fpyrenew-hew+--model_run_dir&type=code)

damonbayer: And let's merge this into main instead of [`add-demo-data`](https://github.com/CDCgov/pyrenew-hew/tree/add-demo-data)

damonbayer: @dylanhmorris suggests this convention:
`filename.py positional_argument --named-argument namedvalue --other-named-argument othernamedvalue`

Which happens to line up with what I suggested, though I did not know that at the time.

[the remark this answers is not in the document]
dylanhmorris: Going forward, let's follow the convention of `snake_case` for `positional_arguments` and `kebab-case` for `--named-arguments`. So here I favor keeping the argument positional but naming it `"model_run_dir"`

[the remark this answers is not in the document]
dylanhmorris: see also https://github.com/CDCgov/pyrenew-hew/pull/131#discussion_r1847044727

SamuelBrand1: > @dylanhmorris suggests this convention: `filename.py positional_argument --named-argument namedvalue --other-named-argument othernamedvalue`
> 
> Which happens to line up with what I suggested, though I did not know that at the time.

This is in a bit of conflict with two bits of the code base atm:
- `iteration_helper` shell scripts are expecting kwargs
- `argparse` in the python scripts is expecting to pass through `kwargs` only.

The second conflict is not detected by `forecast_state.py` because the underlying script functions are call directly.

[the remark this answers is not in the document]
SamuelBrand1: > Why change from a positional argument to a named argument? And why change from `-` to `_`?

Because it was the expectation of the python scripts underlying the shell scripts in `pipelines/iteration_helpers`?

damonbayer: I'm not sure, but @SamuelBrand1 has made changes to the iteration_helper scripts in https://github.com/CDCgov/pyrenew-hew/pull/138, so maybe that resolves it?

SamuelBrand1:
Yeah, and I'll add a small addition to gitignore in #117

PR #135, opened 2024-11-18
Script to run small version of prod pipeline with current branch

damonbayer:
Adds `test` argument to `main` function in `pipelines/batch/setup_prod_job.py`
This argument changes the default mcmc draws number, the locations processed, and storage for results.

This is then used in `pipelines/batch/setup_test_prod_job.py`, which runs the above function with `test=True` using the `pyrenew` container with tag of the current github branch, which is automatically created when that branch has an open PR.

### DIFF
--- a/pipelines/batch/setup_prod_job.py
+++ b/pipelines/batch/setup_prod_job.py
@@ -152,7 +152,7 @@ def main(
 
     all_locations = [
         loc
-        for loc in locations.get_column("STUSAB").to_list()
+        for loc in locations.get_column("STUSAB").to_list() + ["US"]
         if loc not in excluded_locations
     ]
 
--- a/pipelines/batch/setup_prod_job.py
+++ b/pipelines/batch/setup_prod_job.py
@@ -82,8 +82,8 @@ def main(
     pyrenew_hew_output_container = (
         "pyrenew-test-output" if test else "pyrenew-hew-prod-output"
     )
-    n_warmup = 10 if test else 1000
-    n_samples = 10 if test else 500
+    n_warmup = 200 if test else 1000
+    n_samples = 200 if test else 500
 
     creds = EnvCredentialHandler()
     client = get_batch_service_client(creds)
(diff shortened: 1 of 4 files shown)


pipelines/batch/setup_prod_job.py
```diff
@@ -78,6 +79,12 @@ def main(
             f"supported diseases are: {', '.join(supported_diseases)}"
         )
 
+    pyrenew_hew_output_container = (
+        "pyrenew-test-output" if test else "pyrenew-hew-prod-output"
+    )
+    n_warmup = 10 if test else 1000
```
dylanhmorris, line 85: I think we may as well draw enough samples to make results are interpretable, even in test runs. Sampling time unlikely to be the main bottleneck in running this and it makes the test stronger.

pipelines/batch/setup_test_prod_job.py
```diff
@@ -0,0 +1,82 @@
...
+from datetime import datetime, timezone
+from pathlib import Path
+
+from pygit2 import Repository
+from setup_prod_job import main
+
+if __name__ == "__main__":
+    current_datetime = datetime.now(timezone.utc).strftime("%Y%m%d%H%M%SZ")
+    current_branch = Path(Repository(os.getcwd()).head.name).stem
+    locs_to_exclude = [  # keep CA, MN, SD
```
dylanhmorris, line 18: And US, yes?

dylanhmorris (approved):
A few minor suggestions, none critical. Thanks, @damonbayer!

pipelines/batch/setup_test_prod_job.py
```diff
@@ -0,0 +1,82 @@
...
+import os
+from datetime import datetime, timezone
+from pathlib import Path
+
+from pygit2 import Repository
+from setup_prod_job import main
+
+if __name__ == "__main__":
+    current_datetime = datetime.now(timezone.utc).strftime("%Y%m%d%H%M%SZ")
+    current_branch = Path(Repository(os.getcwd()).head.name).stem
```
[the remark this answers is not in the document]
damonbayer: I think I will make the branch name configurable instead.

Issue #126, opened 2024-11-14
Run `forecast_state` end-to-end test on a self-hosted runner in CI using real data


damonbayer:
Pull requests could trigger a light version of the production workflow (only a few mcmc samples for a few states) to verify the pipeline still works.

damonbayer:
The work in https://github.com/CDCgov/pyrenew-hew/pull/135 could be useful for this.

SamuelBrand1:
Also, #148 can be useful for the similar things but covers different part of code base.

SamuelBrand1:
> The work in #135 could be useful for this.
> 
> Aside from integrating into CI, we also need to develop a solution so that multiple PR's on the same data do not have race conditions and overwrite the same files.

Could you make this a bit more precise?

If you're concerned abut the github actions in CI, then so long as each workflow is separate then they should be concurrent by default, so should not interfere at that stage. I don't think committing any results is planned in this issue?

damonbayer:
@SamuelBrand1 when running the pipeline, the results are written to azure blob storage, with the storage directory being named after a small set of data attributes. If you ran a model on the same dataset (as you might if you have multiple PR's in progress) those models would be writing results to the same place (with the same file names).

It's not a terribly difficult problem to solve, but does need to be solved.

damonbayer:
Reworded original description to match https://github.com/CDCgov/pyrenew-hew/issues/216

PR #136, opened 2024-11-18
Add `model_runs` directory

damonbayer:
This makes it easier to iterate over all the runs by calling `dir_ls` on `model_runs` and also makes it easier to locate collated  plots and other files that apply to all the model runs.


damonbayer:
Seems like the baseline forecasts are not being fit.

SamuelBrand1:
> Seems like the baseline forecasts are not being fit.

Are are we making updates to the structure without having a test run facility to check which parts of the code base are affected by any change?

SamuelBrand1:
> @SamuelBrand1 #135 was supposed to help with that. It's not written as a formal test, but that is the plan in #126. Surprisingly, for this PR, it runs without throwing an error, even though the baseline forecasts aren't being fit.

That suggests that we want something to return this failure in part of the run?

damonbayer:
> @SamuelBrand1 #135 was supposed to help with that. It's not written as a formal test, but that is the plan in #126. Surprisingly, for this PR, it runs without throwing an error, even though the baseline forecasts aren't being fit.

This will be addressed in https://github.com/CDCgov/pyrenew-hew/pull/143

damonbayer:
The default "Files Changed" view is borked, but you can see the actual changes [here](https://github.com/CDCgov/pyrenew-hew/compare/main..dmb_model_runs_2).

damonbayer:
@dylanhmorris Ready for your review. Would appreciate some guidance on modification to the collation scripts. You're welcome to adjust them directly on the branch.

pipelines/postprocess_state_forecast.R
```diff
@@ -271,12 +271,15 @@ p <- arg_parser("Generate forecast figures") |>
...
-
-disease_name_raw <- base_dir |>
-  path_file() |>
+# replace this with functionality from hewr
+disease_name_raw <- model_run_dir |>
+  path_split() |>
+  pluck(1) |>
+  tail(3) |>
+  head(1) |>
   str_extract("^.+(?=_r_)")
```
dylanhmorris, line 280: This can happen now that #141 has landed. Use `hewr::parse_model_run_dir_path()`

pipelines/timeseries_forecasts.R
```diff
@@ -68,9 +71,9 @@ fit_and_forecast <- function(data,
                              n_samples = 2000,
                              target_col = "ed_visits",
                              output_col = "other_ed_visits") {
-  forecast_horizon <- glue::glue("{n_forecast_days} days")
-  target_sym <- rlang::sym(target_col)
-  output_sym <- rlang::sym(output_col)
+  forecast_horizon <- glue("{n_forecast_days} days")
+  target_sym <- sym(target_col)
+  output_sym <- sym(output_col)
```
dylanhmorris, line 76: At some point we should make this more consistent with https://dplyr.tidyverse.org/articles/programming.html but it doesn't have to be now.

pipelines/timeseries_forecasts.R
```diff
@@ -250,10 +253,12 @@ disease_name_nssp_map <- c(
...
-base_dir <- path_dir(model_run_dir)
-
-disease_name_raw <- base_dir |>
-  path_file() |>
+# replace this with functionality from hewr
+disease_name_raw <- model_run_dir |>
+  path_split() |>
+  pluck(1) |>
+  tail(3) |>
+  head(1) |>
```
dylanhmorris, line 261: As above, use `hewr` now.

dylanhmorris (changes requested):
One minor change to keep things up to date and then this can land. Thanks @damonbayer!

pipelines/postprocess_state_forecast.R
damonbayer: Resolved in https://github.com/CDCgov/pyrenew-hew/pull/136/commits/3f8139b920dbba75c425e6851be835b47d22f050

pipelines/timeseries_forecasts.R
damonbayer: Resolved in https://github.com/CDCgov/pyrenew-hew/pull/136/commits/3f8139b920dbba75c425e6851be835b47d22f050

pipelines/timeseries_forecasts.R
damonbayer: https://github.com/CDCgov/pyrenew-hew/issues/150

dylanhmorris (approved):
LGTM, thanks @damonbayer!

PR #152, opened 2024-11-20
Add `model_runs` directory (3rd attempt)

damonbayer:
Most of these changes were approved in https://github.com/CDCgov/pyrenew-hew/pull/136. Additional changes begin with https://github.com/CDCgov/pyrenew-hew/commit/26cf4aeca1bc715891c306579edbc5aee4c88675


pipelines/collate_plots.py
```diff
@@ -136,17 +153,24 @@ def process_dir(
         named subdirectories. If ``None``, look in all
         subdirectories of ``base_dir``. Default ``None``.
     """
+    if save_dir is None:
+        save_dir = Path(base_dir).parent / "figures"
```
dylanhmorris, line 157: Create if not exists?

pipelines/collate_plots.py
```diff
@@ -167,10 +190,17 @@ def collate_from_all_subdirs(
...
+        Directory in which to save the merged PDFs.
+        If ``None``, use a "figures" directory in the parent directory of ``model_base_dir``. Default ``None``.
+
     Returns
     -------
     None
     """
+    if save_dir is None:
+        save_dir = Path(model_base_dir).parent / "figures"
+
```
dylanhmorris, line 203: As above, create if not exists?

dylanhmorris (approved):
One minor suggestion, not critical

pipelines/collate_plots.py
damonbayer: This function calls `merge_pdfs_from_subdirs`, which makes the directory if it doesn't exist.

pipelines/collate_plots.py
damonbayer: This function eventually calls `merge_pdfs_from_subdirs`, which makes the directory.

Issue #101, opened 2024-11-04
Reorganize model directories


damonbayer:
This issue refers to the location-level directories. These currently contain a flat list of poorly named files. This is already difficult to navigate and will be increasingly so as we add more fable models and pyrenew models.
I propose having a common `data` and `figures` directory and a separate directory for each model's specific output.

- `data`
- `pyrenew-e`
- `pyrenew-ew`
- `fable`
- `figures`

SamuelBrand1:
For clarification... this is a reorganistion _within_ each of the model definition folders?

SamuelBrand1:
Assuming the above is true, I think the simplest sequence is to do #119 (or some other PR that closes #57) then reorg the internal structure of each location-reportdate directory.

damonbayer:
Started working on this, (see edited original post) but got a bit lost. I think it would be helpful to first develop a data dictionary-like resource about what these files are and what they contain.

damonbayer:
I'm wondering if we should actually put everything here in a `pyrenew-e` directory.

damonbayer:
Updated the initial post with my current thinking.

Issue #79, opened 2024-10-28
epiweekly denominator forecasts


dylanhmorris:
Right now, we forecast both the numerator and the denominator daily and aggregate to epiweekly as needed.

Given that percentage forecasts are likely to be viewed at the epiweekly scale and the denominator is modeled amechanistically, it may be more robust to fit a model of the epiweekly (versus daily) denominator and join in to the numerator forecasts only once they have been aggregated to epiweekly.

PR #159, opened 2024-11-22
Issue 110: Add epiweekly fits and scoring

### LINKED ISSUE #110: Scoring on (epi)weekly scale
At the moment scoring (on the scoring branch) is done daily, and averaged over a week.

This could be useful to keep this, but noting that this includes scoring within-week effects, but we should also have weekly aggregated then scored utility.

### Sub-task planning

- [x] Land #148 
- [x] Add generate epiweekly truth data from daily truth data step
- [x] Add run baseline models on epiweekly
- [x] Add extra pyrenew-hew postprocess to chunk the daily posterior preds into epiweekly
- [x] check how much revision the scoring function needs (might be zero)

SamuelBrand1:
This PR closes #110 .

NB this PR uses `forecasttools` functions, in particular `daily_to_epiweekly`. Very handy.

The goal is to compare the _epiweekly_ forecasts of baseline models (which are fitted to epiweekly data) to the _epiweekly_ grouped forecasts from `pyrenew_hew`.

The daily analysis is unaffected; that is baseline models are still run for daily reporting as well.

## Tasks to contribute
- [x] Add generate epiweekly truth data from daily truth data step
- [x] Add run baseline models on epiweekly
- [x] Add extra pyrenew-hew postprocess to chunk the daily posterior preds into epiweekly
- [x] check how much revision the scoring function needs (might be zero)

## Testing
At each step the test mode pipeline is run to check for failure or breaking changes.


SamuelBrand1:
Added a step to convert daily resolved data to epiweekly cf `generate_epiweekly.R`.

SamuelBrand1:
Modified `timeseries_forecasts.R` to run baseline forecasters for both for daily and epiweekly models.

SamuelBrand1:
Added scoring on epiweekly scale for `pyrenew`. The algorithm is that if `epiweekly = TRUE` then the pyrenew forecasts for counts and other counts are group summed by epiweek (which auto does by draw), pivoted out to create the correct _weekly_ percentage (which is not the mean of the daily percentages), and then pivoted longer.

Dates for the week are supplied by more forecasttools helpers.

No other changes were need because the scoring is flexible enough to ingest this.

pipelines/score_forecast.R
```diff
@@ -141,24 +143,37 @@ prep_truth_data <- function(truth_data_path) {
...
+  if (epiweekly) {
+    message(glue::glue("Scoring epiweekly {model_run_dir}..."))
+  } else {
+    message(glue::glue("Scoring daily  {model_run_dir}..."))
+  }
+  prefix <- ""
+  if (epiweekly) {
+    prefix <- "epiweekly_"
+    eval_data_filename <- glue::glue("{prefix}{eval_data_filename}")
+  }
```
damonbayer, line 162:
```suggestion
  if (epiweekly) {
    eval_data_filename <- glue::glue("epiweekly_{eval_data_filename}")
  }
```

damonbayer:
@SamuelBrand1 Since we are making even more model results and anticipate adding PyRenew-EW to the workflow soon, I would really like to push toward https://github.com/CDCgov/pyrenew-hew/issues/101. Could the fable-related parts of that task be accomplished in this PR (i.e. let's store all the results from these time series models in a `fable` directory)?

pipelines/score_forecast.R
```diff
@@ -169,7 +184,29 @@ read_and_score_location <- function(model_run_dir,
...
+      pivot_wider(names_from = disease, values_from = .value) |>
+      mutate(prop_disease_ed_visits = Disease / (Disease + Other)) |>
+      pivot_longer(c(Disease, Other, prop_disease_ed_visits),
+        names_to = "disease",
+        values_to = ".value"
+      ) |>
+      mutate(date = epiweek_to_date(epiweek, epiyear,
+        day_of_week = day_of_week
+      ))
+  }
```
damonbayer, line 207: Should something like this instead be done in `pipelines/postprocess_state_forecast.R`?

We would actually like to create epiweekly forecasts and present them, right? See https://github.com/CDCgov/pyrenew-hew/issues/79

dylanhmorris: Yes, I think it makes sense to have an epiweekly forecast as canonical output, not just for scoring.

Related: perhaps the location-specific runs should build a location-specific hubverse table, and the overall hubverse table can then just be built via collation (right now it's built via iteration over the results in `create_hubverse_table.R`)

dylanhmorris: I had need of this just now so I did it in 6627518

damonbayer:
> > @SamuelBrand1 Since we are making even more model results and anticipate adding PyRenew-EW to the workflow soon, I would really like to push toward #101. Could the fable-related parts of that issue be accomplished in this PR (i.e. let's store all the results from these time series models in a `fable` directory)?
> 
> I'd rather avoid mission creep on PRs and complete then do another issue.

I don't see the point of adding new code that saves things in locations that we don't like, but I defer to you.

pipelines/batch/setup_prod_job.py
```diff
@@ -1,12 +1,11 @@
 """
-Set up a multi-location, multi-date,
-potentially multi-disease end to end
-retrospective evaluation run for pyrenew-hew
-on Azure Batch.
+Set up a multi-location,  multi-disease production run
```
damonbayer, line 2:
```suggestion
Set up a multi-location, multi-disease production run
```

pipelines/batch/setup_eval_job.py
```diff
@@ -21,6 +22,7 @@ def main(
     job_id: str,
     pool_id: str,
     diseases: str,
+    output_subdir: str | Path = "./",
```
damonbayer, line 25: Why is the output directory not configurable? And if it were, shouldn't we just have one arg where you can put the full path?

pipelines/batch/setup_eval_job.py
```diff
@@ -158,12 +165,17 @@ def main(
...
+        exclude_last_n = 3
         task = get_task_config(
             f"{job_id}-{loc}-{disease}-{report_date}",
             base_call=base_call.format(
                 state=loc,
                 disease=disease,
                 report_date=report_date,
+                n_training=n_training,
+                exclude_last_n=exclude_last_n,
+                output_dir=str(Path("output", output_subdir)),
```
damonbayer, line 178: Should the "output" part be configurable?

pipelines/batch/setup_parameter_inference_job.py
```diff
@@ -0,0 +1,240 @@
...
+        "--state {state} "
+        "--n-training-days 450 "
+        "--n-warmup {n_warmup} "
+        "--n-samples {n_samples} "
+        "--facility-level-nssp-data-dir nssp-etl/gold "
+        "--state-level-nssp-data-dir "
+        "nssp-archival-vintages/gold "
+        "--param-data-dir params "
+        "--output-dir {output_dir} "
+        "--priors-path config/parameter_inference_priors.py "
```
damonbayer, line 142: What are these?

pipelines/batch/setup_parameter_inference_job.py
damonbayer, line 1: What is this job for? Seems like it uses more data and different priors than the usual run?

pipelines/batch/setup_prod_job.py
```diff
@@ -167,6 +174,7 @@ def main(
                 report_date="latest",
                 n_warmup=n_warmup,
                 n_samples=n_samples,
+                output_dir=str(Path("output", output_subdir)),
```
damonbayer, line 177: I have the same concerns about output here as in `setup_eval_job.py`

pipelines/timeseries_forecasts.R
```diff
@@ -142,9 +146,19 @@ cdc_flat_forecast <- function(data,
...
+    model_run_dir, n_forecast_days = 28, n_samples = 2000,
+    epiweekly = FALSE) {
+  prefix <- ""
+  dataname <- "data"
+  data_frequency <- "1 day"
+  if (epiweekly) {
+    dataname <- glue::glue("epiweekly_{dataname}")
+    prefix <- "epiweekly_"
+    data_frequency <- "1 week"
+  }
```
damonbayer, line 159: I think either
```r
if (epiweekly) {
  # Assign variables
} else {
  # Assign variables
}
```
or separate
```r
prefix <- if_else(epiweekly, "epiweekly_", "")
```
is easier to read

pipelines/score_hubverse.R
```diff
@@ -0,0 +1,85 @@
...
+
+truth_path <- "~/epiweekly.tsv"
+all_paths_flu <- c(
+  "~/2024-11-04-influenza-hubverse-table.tsv",
+  "~/2024-11-13-influenza-hubverse-table.tsv"
+)
+all_paths_covid <- c(
+  "~/2024-11-04-covid-hubverse-table.tsv",
+  "~/2024-11-13-covid-19-hubverse-table.tsv"
+)
```
damonbayer, line 55: can these assigned programmatically (`dir_ls`?)

pipelines/score_hubverse.R
```diff
@@ -0,0 +1,85 @@
...
+
+flu_scores <- purrr::map(
+  all_paths_flu,
+  \(x) score_hubverse(truth_path, x, prop_influenza)
+) |>
+  dplyr::bind_rows() |>
+  dplyr::filter(target_end_date < lubridate::today() + 2)
```
damonbayer, line 62: why?

pipelines/score_hubverse.R
```diff
@@ -0,0 +1,85 @@
...
+covid_scores <- purrr::map(
+  all_paths_covid,
+  \(x) score_hubverse(truth_path, x, prop_covid)
+) |>
+  dplyr::bind_rows() |>
+  dplyr::filter(target_end_date < lubridate::today() + 2)
```
damonbayer, line 68: why?

pipelines/score_hubverse.R
```diff
@@ -0,0 +1,85 @@
...
+full_scores <- dplyr::bind_rows(
+  flu_scores,
+  covid_scores
+)
+
+
+flu_summary <- flu_scores |>
+  scoringutils::summarise_scores(by = c("horizon", "reference_date", "target"))
+
+covid_summary <- covid_scores |>
```
damonbayer, line 79: Overall seems like you could `walk` or loop though each disease and use a common function.

pipelines/score_hubverse.R
```diff
@@ -0,0 +1,85 @@
...
+  scoringutils::summarise_scores(by = c("horizon", "reference_date", "target"))
+
+
+full_summary <- dplyr::bind_rows(flu_summary, covid_summary)
+
+fig <- plot_coverage_by_date(full_scores, 0.95)
```
damonbayer, line 85: What do you do with this figure? Should it be saved somewhere?

pipelines/create_observed_data_tables.py
damonbayer, line 1: General comment that the pipelines folder is becoming unwieldy. After this PR, I count 21 scripts. Additional organization would be helpful.

pipelines/default_priors.py
damonbayer, line 1: Why is the filemode changing?

damonbayer, line 1: What is motivating the changes to the priors?

pipelines/generate_epiweekly.R
```diff
@@ -0,0 +1,96 @@
...
+#' @param strict A logical value indicating whether to enforce strict inclusion
+#' of only full epiweeks. Default is TRUE.
+#' @param day_of_week An integer specifying the day of the week to use for the
+#' epiweek date. Default is 1 (Monday).
+#'
+#' @return None. The function writes the epiweekly data to a CSV file in the
+#'  specified directory.
+convert_daily_to_epiweekly <- function(
+    model_run_dir, strict = TRUE,
+    day_of_week = 7, dataname = "data", ext = "csv") {
```
damonbayer, line 33: I think it makes more sense to take the `data_path` as the arg and use utilities from `fs` to extract the parts as necessary.

pipelines/plot_parameters.R
damonbayer, line 1: Does this plot get saved somewhere?

damonbayer (commented):
Nothing that would necessarily prevent a merge, but I had lots of comments and questions.

pipelines/batch/setup_eval_job.py
dylanhmorris: No, because that's the hard-coded bind mount location of the output blob storage container within the running podman/docker container.

pipelines/batch/setup_eval_job.py
```diff
@@ -108,7 +115,7 @@ def main(
                 "target": "/pyrenew-hew/params",
             },
             {
-                "source": "pyrenew-hew-prod-output",
+                "source": "pyrenew-test-output",
                 "target": "/pyrenew-hew/output",
```
dylanhmorris, line 119: Noting the target of `output`

pipelines/batch/setup_eval_job.py
dylanhmorris: See https://github.com/CDCgov/pyrenew-hew/pull/159/files/a2fa9afa9651fda572c50b4a2db90b8379dc468b#r1859432678

pipelines/batch/setup_eval_job.py
dylanhmorris: See https://github.com/CDCgov/pyrenew-hew/pull/159/files/a2fa9afa9651fda572c50b4a2db90b8379dc468b

Because this is just an alias for the root of the output blob storage container, I don't think it needs a configurable name.

pipelines/batch/setup_parameter_inference_job.py
dylanhmorris: Yes, idea is to use this script for runs with longer timeseries and less informative priors. This might be useful, e.g., for choosing relatively informative priors for shorter timeseries fits.

pipelines/batch/setup_parameter_inference_job.py
dylanhmorris: They live in the `pyrenew-hew-config` blob storage container.

pipelines/batch/setup_prod_job.py
dylanhmorris: See response above.

pipelines/create_observed_data_tables.py
dylanhmorris: Agree but out of scope here. And this one is a rename rather than a new file.

pipelines/default_priors.py
dylanhmorris: Guessing this or something like it? https://stackoverflow.com/questions/1257592/how-do-i-remove-files-saying-old-mode-100755-new-mode-100644-from-unstaged-cha

pipelines/batch/setup_parameter_inference_job.py
damonbayer: Should they be reviewed as part of code review?

pipelines/score_hubverse.R
damonbayer: Should the reference date be computed based on the data, rather than the time the script is run?

pipelines/timeseries_forecasts.R
SamuelBrand1: Agreed. Much nicer. I've changed this now.

pipelines/batch/setup_eval_job.py
damonbayer: Why are you noting the target of `output`? Are you uncertain about it?

damonbayer (approved):
@dylanhmorris This PR has my tentative approval. Please respond to (or make changes based on) the remaining comments before merging.

pipelines/generate_epiweekly.R
SamuelBrand1: I left `dataname` as an arg, because `model_run_dir` is a common arg across multiple processes (although obvs we could change this in the associated `main` function so feel free to come back at me on that). 

However, I agree than the `ext` doesn't need to be a separate arg!

pipelines/batch/setup_parameter_inference_job.py
dylanhmorris: This is a good question about practice for setting configuration and how much this repo is implementation versus configuration, but imo out of scope for this PR. Will open an issue.

pipelines/score_hubverse.R
dylanhmorris: This file shouldn't really be in this PR at all. Going to remove from this PR and merge it to separately, cleaned up and with your (good) comments addressed

dylanhmorris: This file shouldn't really be in this PR at all. Going to remove from this PR and merge it to separately, cleaned up and with your comments addressed

pipelines/batch/setup_eval_job.py
dylanhmorris: So I could link to it from the other comments.

PR #188, opened 2024-12-05
Improve postprocessing for retro evaluation results

dylanhmorris:
- Simplifies and DRYifies relative wis computation
- Renames functions to be more meaningful
- DRYifies creation of tables


hewr/R/score_hubverse.R
```diff
@@ -87,7 +88,7 @@ score_hubverse <- function(forecast,
     scoringutils::score(
       metrics = c(
         scoringutils::get_metrics(to_score),
-        list(interval_coverage_95 = interval_coverage_95)
+        interval_coverage_95 = interval_coverage_95
```
damonbayer, line 91: Did this fail before?

dylanhmorris: No, it just struck me that since `get_metrics` returns a list this would be less verbose

pipelines/postprocess_scoring.R
```diff
@@ -39,139 +39,121 @@ purrr::walk(script_packages, \(pkg) {
...
+                                date_column = "date",
+                                score_column = "relative_wis",
+                                model_column = "model",
+                                plot_title = "Scores by model over time",
+                                xlabel = "Date",
+                                ylabel = "Relative WIS") {
+  min_score <- min(scores_by_date[[score_column]])
+  max_score <- max(scores_by_date[[score_column]])
+  max_overall <- max(c(1 / min_score, max_score))
+  sym_ylim <- c(1 / max_overall, max_overall)
```
damonbayer, line 96: you do this a few times, so there is an opportunity to DRY by creating a `compute_sym_lim` function

dylanhmorris: Yes, I was thinking of dumping it into `forecasttools`

pipelines/postprocess_scoring.R
```diff
@@ -329,54 +366,41 @@ main <- function(path_to_scores,
...
-
-  table_locs_save_path <- get_save_path(
-    "scores_by_location",
-    ext = "tsv"
-  )
-  readr::write_tsv(table_locs, table_locs_save_path)
-
-  message("Done with score postprocessing.")
+  message("Done with scoring postprocessing.")
 }
```
damonbayer, line 404: For saving a bunch of figures, I like the pattern of building a table with the figures, file names, dimensions, etc and  `pwalk`ing, but this is fine.

damonbayer (approved):
You can merge, but please make issues for any of my comments that you think should be addressed in the future.

PR #193, opened 2024-12-06
DRAFT: Move models into separate directories

damonbayer:

PR #197, opened 2024-12-09
Update to_epiweekly_quantile_table to work with new output schema

dylanhmorris:


dylanhmorris:
Separately will add a test that would've caught this, but want main to work in the meantime.

damonbayer (approved):
Sorry I missed this in https://github.com/CDCgov/pyrenew-hew/pull/193.

PR #200, opened 2024-12-09
Create pull_nhsn.R

damonbayer:
This script adds a `pull_nhsn.R` script to the `pipelines` folder. A future PR will integrate this script into the pipeline, once we are ready to work on `H` models.

### DIFF
--- a/pipelines/pull_nhsn.R
+++ b/pipelines/pull_nhsn.R
@@ -2,6 +2,7 @@ script_packages <- c(
   "forecasttools",
   "dplyr",
   "readr",
+  "lubridate",
   "argparser"
 )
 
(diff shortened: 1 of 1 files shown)


pipelines/pull_nhsn.R
```diff
@@ -0,0 +1,66 @@
+script_packages <- c(
+  "forecasttools",
+  "tidyverse",
```
dylanhmorris, line 3:
```suggestion
  "dplyr",
  "readr",
```
Prefer not to load the whole tidyverse but rather just the packages used

dylanhmorris (approved):
One minor suggestion

PR #206, opened 2024-12-10
Modularize `process_state_forecast.R` and add some basic tests

dylanhmorris:
Prerequisite for epiweekly predicted/actual plots but separable and therefore merging in now to prevent mega PRs


hewr/R/process_state_forecast.R
```diff
@@ -1,21 +1,104 @@
...
+#' @param disease_name Name of the target disease.
+#' One of `"COVID-19"` or `"Influenza"`.
+#' @return The combined data, as a [`tibble`][tibble::tibble()].
+#' @export
+combine_training_and_eval_data <- function(train_dat,
+                                           eval_dat,
+                                           disease_name) {
+  combined_dat <- dplyr::bind_rows(
+    train_dat |>
+      dplyr::filter(.data$data_type == "train"),
```
damonbayer, line 15: Now that we have sorted out a way to get eval data, we should probably stop having non-training data in the `train_dat`. Would require changes to https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/prep_data.py

hewr/R/process_state_forecast.R
```diff
@@ -30,135 +113,119 @@ process_state_forecast <- function(model_run_dir,
...
+  )
 
   last_training_date <- combined_dat |>
-    dplyr::filter(data_type == "train") |>
-    dplyr::pull(date) |>
+    dplyr::filter(.data$data_type == "train") |>
+    dplyr::pull("date") |>
     max()
 
   other_ed_visits_samples <-
```
damonbayer, line 134: I know I wrote this, but I think this part could use a comment to explain that we are creating a fake posterior by replicating the true data.

damonbayer (approved):
Just a few suggestions. Please address directly or make issues before merging.

hewr/R/process_state_forecast.R
dylanhmorris: Created #207

hewr/R/process_state_forecast.R
dylanhmorris: done in e1ff3d7

PR #208, opened 2024-12-11
Update pre-commit.yaml to use pre-commit/action

dylanhmorris:


damonbayer:
@sbidari Does this solve your issue with pre-commit?

damonbayer (approved):

PR #215, opened 2024-12-11
Fixing the tag in container delete

gvegayon:
It was using the wrong tag name.


.github/workflows/delete-container-tag.yaml
```diff
@@ -23,14 +23,23 @@ jobs:
...
+
       - name: Login to the Container Registry
         uses: docker/login-action@v3
         with:
           registry: "cfaprdbatchcr.azurecr.io"
           username: "cfaprdbatchcr"
           password: ${{ secrets.CFAPRDBATCHCR_REGISTRY_PASSWORD }}
 
-      - name: Login to Azure with STF Service Principal
+      - name: Login to Azure with NNH Service Principal
```
gvegayon, line 42:
```suggestion
      - name: Login to Azure with STF Service Principal
```

.github/workflows/delete-container-tag.yaml
```diff
@@ -43,4 +52,5 @@ jobs:
         run: |
           az acr repository delete \
             --name ${{ env.REGISTRY }} \
-            --image ${{ env.IMAGE_NAME }}:${{ steps.branch-name.outputs.branch}}
+            --image ${{ env.IMAGE_NAME }}:${{ steps.image-tag.outputs.tag}}
```
gvegayon, line 55:
```suggestion
            --image ${{ env.IMAGE_NAME }}:${{ steps.image-tag.outputs.tag }}
```

PR #222, opened 2024-12-11
Predicted/actual epiweekly plots

dylanhmorris:
- Adds functionality to create multiple visualizations of predicted versus actual values from hubverse tables
- Minor tweaks to other postprocessing.


damonbayer:
These will actually be generated automatically in https://github.com/CDCgov/pyrenew-hew/issues/287

damonbayer:
> These will actually be generated automatically in #287

This is not true. I misunderstood the goal of this PR.

pipelines/collate_score_tables.R
```diff
@@ -81,6 +81,29 @@ bind_tables <- function(list_of_table_pairs) {
   ))
 }
 
+save_scores <- function(score_table,
+                        score_file_save_path = NULL) {
+  if (!is.null(score_file_save_path)) {
+    message(glue::glue(paste0(
+      "Saving score table to ",
+      "{score_file_save_path}..."
+    )))
```
damonbayer, line 90: paste is redundant. Glue will concatenate the strings on its own.

pipelines/collate_score_tables.R
```diff
@@ -102,18 +125,36 @@ collate_scores_for_date <- function(model_run_dir,
...
+    NULL
   }
+
+  save_scores(date_score_table, save_path)
+
   message(glue::glue("Done processing scores for {model_run_dir}."))
   return(date_score_table)
 }
 
+collate_from_dirs <- function(dirs,
```
damonbayer, line 144: I don't see this function being used anywhere?

dylanhmorris: It can be removed. It was a part of the unpruned version that I forgot to prune, and I think we'll refactor scoring considerably now that we have the new postprocessing.

dylanhmorris: Done in b14dbbf

pipelines/collate_score_tables.R
dylanhmorris: Done in b14dbbf

pipelines/collate_score_tables.R
```diff
@@ -81,6 +81,18 @@ bind_tables <- function(list_of_table_pairs) {
...
+save_scores <- function(score_table,
+                        score_file_save_path = NULL) {
+  if (!is.null(score_file_save_path)) {
+    message(glue::glue(
+      "Saving score table to ",
+      "{score_file_save_path}..."
+    ))
+    readr::write_rds(score_table, score_file_save_path)
+  }
+}
```
damonbayer, line 93: Is this function necessary? Do we expect to call it with a NULL `score_file_save_path`?

dylanhmorris: I think it's a vestige of a pruned workflow. Removed in 014651e

damonbayer:
Per F2F conversation, @dylanhmorris is going to clean up this PR and re-request review when ready.

dylanhmorris:
It should be ready for re-review now.

pipelines/hubverse_score.R
```diff
@@ -15,6 +18,84 @@ get_hubverse_table_paths <- function(dir,
...
+      ),
+      color = "blue",
+      point_fill = "darkblue",
+      point_color = "black",
+      point_alpha = 1,
+      alpha = 0.75,
+      linewidth = 20,
+      point_size = 4,
+      shape = 23
+    ) +
```
damonbayer, line 69: can't this be achieved with a single `geom_pointinterval` call?

pipelines/hubverse_score.R
```diff
@@ -15,6 +18,84 @@ get_hubverse_table_paths <- function(dir,
...
+      ),
+      names_from = quantile_level,
+      names_glue = "q_{quantile_level * 100}",
+      values_from = predicted
+    )
+
+  plot <- to_plot |>
+    ggplot(aes(
+      x = target_end_date,
+      y = q_50
```
damonbayer, line 44: Suggest moving this into the `geom_pointinterval` call, since it is overwritten in the later geoms.

pipelines/hubverse_score.R
```diff
@@ -110,19 +204,136 @@ score_and_save <- function(observed_data_path,
...
+            dplyr::filter(
+              .data$horizon %in% !!horizons
+            ),
+          x
+        )
+      }
+    )
+
+  pred_act_plot_fn <- function(location, disease) {
+    all_horizons <- unique(horizons)
```
damonbayer, line 240: Why this now?

pipelines/hubverse_score.R
```diff
@@ -110,19 +204,136 @@ score_and_save <- function(observed_data_path,
...
+      ) |>
+      dplyr::anti_join(loc_table,
+        by = c(
+          "reference_date",
+          "horizon",
+          "target_end_date"
+        )
+      )
+
+    synth_minus_one <- dplyr::inner_join(needed,
```
damonbayer, line 276: What is meant by `synth_minus_one`?

pipelines/collate_score_tables.R
damonbayer, line 1: Can't comment on these lines, but I don't like the separation of `score_file_name`, `score_file_ext`, and 
 `score_file_save_path` as arguments. I think better for the user to supply a single file path. They can construct it with `fs::path` outside of the function, if they want to.

pipelines/hubverse_score.R
dylanhmorris: Potentially but when as here there are pre-calculated CI bounds I find this more readable and not that much longer.

pipelines/hubverse_score.R
dylanhmorris: Done in b0631e9

pipelines/hubverse_score.R
dylanhmorris: Left over from an abandoned approach when the `full_scorable_table` needed not to be filtered by horizon immediately. Simplified in 0984d2c

pipelines/hubverse_score.R
dylanhmorris: Added explanatory comment in 0984d2c

pipelines/collate_score_tables.R
```diff
@@ -147,44 +154,74 @@ collate_all_score_tables <- function(model_base_dir,
...
   return(full_score_table)
 }
 
+main <- function(dir_of_forecast_dirs,
+                 diseases = c("COVID-19", "Influenza"),
+                 score_file_names = c(
+                   "score_table",
+                   "epiweekly_score_table"
+                 ),
+                 score_file_ext = "rds",
```
damonbayer, line 170: If there's no way to save something other than rds, no need to have this as an argument.

dylanhmorris: case-sensitivity and future-proofing, but happy to simplify

pipelines/hubverse_score.R
damonbayer, line 1: Again can't comment on these lines, but `influenza_table_dir` and `covid_table_dir` as objects does not seem very forward looking. Would rather `dir_ls` and look for any files ending in `-hubverse-table.tsv`

damonbayer: Also, are we planning to have both `tsv` and `parquet` hubverse files?

dylanhmorris: csv/tsv for submittable tables, `.parquet` for archival ones? But perhaps we should reconsider that and/or keep copies of even the submittable tables as parquet right up to submission and then convert when we copy them to hubs?

pipelines/hubverse_score.R
```diff
@@ -15,6 +18,85 @@ get_hubverse_table_paths <- function(dir,
 }
 
 
+plot_pred_act_by_horizon <- function(scorable_table,
+                                     location) {
+  to_plot <- scorable_table |>
+    dplyr::filter(
+      location == !!location,
+      quantile_level %in% c(0.025, 0.25, 0.5, 0.75, 0.975, NA)
```
damonbayer, line 26: should this be a function argument? What if not all the desired quantiles are present?

pipelines/hubverse_score.R
```diff
@@ -15,6 +18,85 @@ get_hubverse_table_paths <- function(dir,
...
+    ) +
+    geom_line(aes(y = observed),
+      size = 1
+    ) +
+    geom_point(aes(y = observed),
+      size = 4,
+      shape = 21,
+      fill = "darkred",
+      color = "black",
+      alpha = 1
```
damonbayer, line 79: can omit

pipelines/hubverse_score.R
```diff
@@ -15,6 +18,85 @@ get_hubverse_table_paths <- function(dir,
...
+    labs(
+      title =
+        glue::glue("Predictions and observations for {location}"),
+      x = "Target date",
+      y = "%ED visits"
+    ) +
+    scale_y_continuous(
+      labels = scales::label_percent(),
+      transform = "log10"
+    ) +
```
damonbayer, line 93: Extreme nit, but I don't like to mix `scale_` and `labs` since you can specify the names in both.

pipelines/hubverse_score.R
```diff
@@ -110,19 +205,126 @@ score_and_save <- function(observed_data_path,
...
+
+  wis_by_loc <- scoringutils::plot_wis(
+    summaries$summary_by_location |>
+      dplyr::arrange(horizon, target, wis) |>
+      dplyr::mutate(location = factor(
+        location,
+        levels = unique(
+          location
+        ),
+        ordered = TRUE
```
damonbayer, line 321: Why specify `levels` and why is it `ordered`?

pipelines/hubverse_score.R
```diff
@@ -68,29 +150,42 @@ score_and_save <- function(observed_data_path,
...
   }
 
-  full_scores <- all_paths |>
-    purrr::pmap(read_and_score) |>
-    dplyr::bind_rows()
+  full_scorable_table <- all_paths |>
+    purrr::pmap(read_and_prep_for_scoring) |>
+    dplyr::bind_rows() |>
+    dplyr::select(
+      -"other_ed_visit_forecast",
```
damonbayer, line 177: I surprised this is a column at this point. Seems like the output of `read_and_prep_for_scoring` is not pivoted correctly to be easily scorable.

pipelines/hubverse_score.R
dylanhmorris: So that locations are ordered from worst to best WIS for the first horizon/target pair, rather than alphabetical.

pipelines/hubverse_score.R
damonbayer: > `csv`/`tsv` for submittable tables, `.parquet` for archival ones?
That sounds good. We should be very clear about what kinds of tables these functions are expected to work with (archival vs submittable)

pipelines/hubverse_score.R
damonbayer: I mean only the `alpha=1` part

pipelines/hubverse_score.R
dylanhmorris: Generalized the "which disease" logic in c7dbe8e

pipelines/hubverse_score.R
dylanhmorris: Right now the function only does 50\% and 95\%, so it makes sense to hard-code. We can generalize later (e.g. if/when we add the function to forecasttools) but I think it's out of scope here.

pipelines/hubverse_score.R
dylanhmorris: Done in 4b5ccae

pipelines/hubverse_score.R
damonbayer: [`forcats::fct_reorder`](https://forcats.tidyverse.org/reference/fct_reorder.html) to make this more clear?

pipelines/hubverse_score.R
dylanhmorris: I see your point, but in this case I think it makes sense because we have an x label and a title but no explicit `scale_x` call. I do think it makes more sense to have the label call after the scale call. I have done this in 679e77e

pipelines/hubverse_score.R
dylanhmorris: The column name is confusing; it's an old (no longer used) column name for recording whether we used an aggregated daily other ed visit forecast or explicitly epiweekly other ed visit forecast for the denominator. Perhaps clearer just to have the columns we want affirmatively selected.

dylanhmorris: Done in ece90dc

pipelines/collate_score_tables.R
dylanhmorris: I think our scoring is going to change considerably as we move to the new output/postprocessing structure, so I'm reluctant to do too much here now.

damonbayer (approved):
Thanks @dylanhmorris! Please fix failing CI before merging.

dylanhmorris:
> Thanks @dylanhmorris! Please fix failing CI before merging.

The CI is the `epipredict` issue documented here: https://github.com/cmu-delphi/epipredict/issues/438. I think we should pin to an older version in a separate PR while a fix is implemented.

PR #230, opened 2024-12-12
Add epiweekly "other" option to `create_hubverse_table.R`

dylanhmorris:


damonbayer (approved):
Looks good, assuming you are running this from the command line, not some other pipeline script.

dylanhmorris:
Correct, though eventually it will/should end up in a pipeline

Issue #227, opened 2024-12-11
Add "Prev" and "Next" buttons to Dashboard and default to viewing forecast plots


damonbayer:
This will mostly remove the need for the joint pdf, which we could remove from the repo.

PR #231, opened 2024-12-12
Dashboard Navigation and Plot Improvements

damonbayer:
- [x] Adds "Prev" and "Next" buttons to navigation to easily navigate between states in an interactive review.
- [ ] Makes forecast page the default page when navigating to a state.
- [x] Adds forecast period to Rt figures.

Closes https://github.com/CDCgov/pyrenew-hew/issues/227
Closes https://github.com/CDCgov/pyrenew-hew/issues/228


damonbayer:
There is some strange bug with quarto where the first page has the wrong layout until you navigate away(https://github.com/quarto-dev/quarto-cli/issues/11672). Until this is resolved, there is no benefit to making the "forecasts" page the first page.

PR #239, opened 2024-12-13
fix for using tag in test job

damonbayer:
Fixes two bugs with test job:

- [ ] No longer overwrites user-specified tag with branch name
- [ ] maps "main" branch to "latest" tag


sbidari (approved):
LGTM

Issue #237, opened 2024-12-13
Track git commit id when fitting models


damonbayer:

damonbayer:
Unfortunately, it doesn't seem like the git commit id is available from within a container run. Perhaps there is a way to modify the container build to make it accessible? @dylanhmorris

PR #242, opened 2024-12-13
Track git commit id when fitting models

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/237

This PR enables us to track the git commit id and branch name of every model fit.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -4,6 +4,9 @@ FROM cfaprdbatchcr.azurecr.io/pyrenew-hew-dependencies:${TAG}
 ARG GIT_COMMIT_SHA
 ENV GIT_COMMIT_SHA=$GIT_COMMIT_SHA
 
+ARG GIT_BRANCH_NAME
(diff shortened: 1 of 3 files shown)


Containerfile
```diff
@@ -1,6 +1,11 @@
 ARG TAG=latest
 
 FROM cfaprdbatchcr.azurecr.io/pyrenew-hew-dependencies:${TAG}
+ARG GIT_COMMIT_SHA
```
dylanhmorris, line 4: should this and the branch name be arguments to the container build call or should the container build engine itself invoke `git` to determine it? The latter seems more robust: will also work with logal builds, will reduce GH actions magic, and it's a pretty minimal dep for container building.

damonbayer:
Per Teams conversation with @dylanhmorris , this can be merged now that https://github.com/CDCgov/pyrenew-hew/issues/249 exists.

PR #245, opened 2024-12-14
Adding the twostep container action from CDCgov/cfa-actions

gvegayon:
It replaces the container workflow with the action from <https://github.com/CDCgov/cfa-actions/>.

### DIFF
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -17,105 +17,18 @@ jobs:
     runs-on: cfa-cdcgov
     name: Build dependencies image
 
-    outputs:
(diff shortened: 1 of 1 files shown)


damonbayer:
I am actually in the middle of modifying the actions here: https://github.com/CDCgov/pyrenew-hew/pull/242

Would I be able to achieve my goal with the CFA actions?

gvegayon:
Build args are currently hard-coded in the action, but adding that as an option would be awesome. I don't think it will be hard to do. I will add the feature and ping you when it's ready. I can start now.

dylanhmorris:
Going to close this in favor of a new PR once https://github.com/CDCgov/cfa-actions/pull/11 lands

PR #246, opened 2024-12-16
Add `pipelines` to codecov 

sbidari:


damonbayer:
Can we split it up so that we have separate coverage for `hewr`, `pyrenew_hew` and `pipelines`? And perhaps use those as the flags names: 
CleanShot 2024-12-16 at 11 02 20@2x

sbidari:
@damonbayer Made the changes, I think `python-tests` and `r-tests` will stop appearing for PRs opened after this is merged. 
image

damonbayer (approved):
Thanks @sbidari!

PR #172, opened 2024-12-03
Draft: Create test pipeline CI - do not merge

SamuelBrand1:
Testing out a CI which installs pyrenew-hew and hewr before running the test pipeline to address #155

PR #247, opened 2024-12-16
Issue 216: end-to-end CI

### LINKED ISSUE #216: Run `forecast_state` end-to-end test on a GitHub-hosted runner in CI
We should run `main` from [forecast_state.py](https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/forecast_state.py) on a GitHub-hosted runner in CI.

It would be great if the files created from this workflow were easily downloadable (especially the resulting dashboard page).

This function expects to be passed `facility_level_nssp_data_dir` and `state_level_nssp_data_dir` and then generates data to be used for model fitting.

The existing test data in https://github.com/CDCgov/pyrenew-hew/tree/main/pipelines/tests/covid-19_r_2024-01-29_f_2023-11-01_t_2024-01-29/model_runs/TD/data is one step too "processed." This level of data should be generated during the test.

I think this may need to be broken up into smaller issues and PR's to keep things manageable for reviewers.

SamuelBrand1:
This PR closes #216.

Steps:

- [x] `hewr` function to generate count data from times series of exponential rates $r_t$.
- [x] `pipeline` function to use above to create dummy data with this schema https://gist.github.com/damonbayer/63bfeab36514806896cd06dc0235d1ea @damonbayer 
- [x] gh actions basically as per (closed) #172 

This PR adds a CI which installs relevant deps, then runs a test script which first creates fake test data from an exponentially mean-varying Poisson process and then runs the forecasting pipeline script. The target pathogen is covid-19 and the target state is "CA" (chosen arbitrarily).


SamuelBrand1:
Blocked by an error 

> raise ValueError("Data frame appears to have date gaps")

Which is unexpected because the data is generated in sequence.

I'll parse this tomorrow.

SamuelBrand1:
That error was due to differing names of disease across different data sets. Onto creating eval data.

SamuelBrand1:
This seems to be running but noting that without the `__pycache__/` dir it takes a while to compile.

SamuelBrand1:
I've tried more "minimalistic" settings e.g. fewer look back points, but occasionally you get fails due to bad sampling propagating something unexpected along the pipeline. At this moment I'm using 90 days data for inference and drawing a single 500/500 chain. 

Exploring the fastest option that 1) doesn't fail stochastically, 2) still covers the pipeline sufficiently to avoid false CI passes would be a good issue.

SamuelBrand1:
I think this is now erroring because `epipredict` and `epiprocess` are not loaded.

damonbayer (commented):
@SamuelBrand1 After I added the quarto action, all checks are passing. On Teams you alluded to a potential other issue, so holding off on approval for now.

SamuelBrand1:
> @SamuelBrand1 After I added the quarto action, all checks are passing. On Teams you alluded to a potential other issue, so holding off on approval for now.

Nice! Thanks @damonbayer . The other issue was using deps that need to be installed with `pak` e.g. `epipredict`; I hadn't got to the Quarto problem yet.

This should be good to review now.

damonbayer (approved):
Thanks @SamuelBrand1. This is a huge improvement for the repo!

PR #248, opened 2024-12-16
Make `setup_prod_job.py` more configurable

dylanhmorris:
To facilitate re-runs of specific locations as needed.

- Adds a `--locations-include` argument
- Renames `--excluded-locations` to `--locations-exclude` for consistency
- Makes `n_training_days` and `exclude_last_n_days` configurable from the command line, with defaults equal to the previously hard-coded values.

### DIFF
--- a/pipelines/batch/setup_prod_job.py
+++ b/pipelines/batch/setup_prod_job.py
@@ -22,7 +22,10 @@ def main(
     output_subdir: str | Path = "./",
     container_image_name: str = "pyrenew-hew",
     container_image_version: str = "latest",
-    excluded_locations: list[str] = [
+    n_training_days: int = 90,
+    exclude_last_n_days: int = 1,
+    locations_include: list[str] = None,
+    locations_exclude: list[str] = [
         "AS",
         "GU",
         "MO",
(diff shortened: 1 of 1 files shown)


pipelines/batch/setup_prod_job.py
```diff
@@ -242,5 +305,7 @@ def main(
 if __name__ == "__main__":
     args = parser.parse_args()
     args.diseases = args.diseases.split()
-    args.excluded_locations = args.excluded_locations.split()
+    if args.locations_include is not None:
+        args.locations_include = args.locations_include.split()
+    args.locations_exclude = args.locations_exclude.split()
```
damonbayer, line 310: Seems like we should either have the `is not None` check for both or neither, unless I am missing something.

dylanhmorris: `--locations-exclude` defaults to a string, not`None`.

dylanhmorris: (and if it is explicitly set to the empty string that will work, as it will result in an empty list)

damonbayer (approved):

Issue #253, opened 2024-12-17
Pipelines Cleanup


damonbayer:
Given the lack of progress on https://github.com/CDCgov/pyrenew-hew/issues/48, I think we are overdue for a cleanup of the pipelines directory.

We should

- [ ] Put scripts into folders
  - [ ] Pre-processing
  - [ ] Model fitting
  - [ ] Post-Processing
  - [ ] Collation
- [ ] Rename files that are specific to certain models to communicate that in the file name
- [ ] Add some sort of header comment in each file that briefly describes what each file does

This should be done in concert with https://github.com/CDCgov/pyrenew-hew/issues/235
- [ ] add a readme

dylanhmorris:
>  Add some sort of header comment in each file that briefly describes what each file does
For Python this can and should be the script docstring (before imports, after shebang if any):
```python
!#/usr/bin/env python3 
"""
What this script does
"""

import mylibrary as ml
```

For R, roxygen-style comment at the top?
```
#' This script does does things
```

PR #254, opened 2024-12-17
Pipelines Cleanup

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/253.


damonbayer:
This is working locally but broken on azure.

damonbayer:
I could not get this to work in azure, so I have reduced the scope to simple file renames.

PR #261, opened 2024-12-19
Add NHSN data pull to forecast_state

damonbayer:
Adds NHSN data pull to `process_and_save_state`. This in turn modifies `data_for_model_fit`. The new `data_for_model_fit` is backwards compatible with existing `pyrenew-e` models and has the necessary data to fit a `pyrenew-he` model.

We also save `combined_training_dat.tsv` which is necessary for post-processing `pyrenew-he` models. We contintue to save `data.tsv` to maintain compatibility with current `pyrenew-e` models.


dylanhmorris (approved):
LGTM. Nice catch on the no-longer-used argument that was hanging around.

PR #265, opened 2024-12-23
Update pyrenew_hew_model sample signature for PyRenew-HE

damonbayer:


damonbayer:
@dylanhmorris is this sufficient for your needs? There are many many places throughout the pipeline that eventually need to be changed to accommodate the new model, but this seems like a reasonable place to start that doesn't impact the pipeline.

dylanhmorris (approved):

PR #266, opened 2024-12-23
hotfix `prep_data.py`

dylanhmorris:
reading a tsv from stdout is not robust to other stuff that might end up in stdout.

EDIT: made more robust with an explicit temp file (parquet format).

### DIFF
--- a/pipelines/prep_data.py
+++ b/pipelines/prep_data.py
@@ -57,9 +57,10 @@ def get_nhsn(
         r_command,
         capture_output=True,
     )
+
     if result.returncode != 0:
         raise RuntimeError(f"pull_and_save_nhsn: {result.stderr}")
-    raw_dat = pl.read_csv(result.stdout, separator="\t")
+    raw_dat = pl.read_csv(result.stdout, skip_rows=1, separator="\t")
     dat = raw_dat.with_columns(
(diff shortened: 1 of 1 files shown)


damonbayer:
@dylanhmorris I think both your proposed solution and the current solution are too dependent on random things that could be printed to stdout. Probably more robust to write to a tmp file from R and read and delete that file in python.

dylanhmorris:
> @dylanhmorris I think both your proposed solution and the current solution are too dependent on random things that could be printed to stdout. Probably more robust to write to a tmp file from R and read and delete that file in python.

I agree and have implemented this.

pipelines/prep_data.py
```diff
@@ -37,6 +42,8 @@ def py_scalar_to_r_scalar(py_scalar):
 
     state_abb = state_abb if state_abb != "US" else "USA"
 
+    temp_file = Path(temp_dir, "nhsn_temp.parquet")
```
damonbayer, line 45: Do we need to check if this exists already and error?

dylanhmorris: because we explicitly created the enclosing `temp_dir` above via https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp this should be safe without that check

PR #276, opened 2025-01-07
Remove time column from output

damonbayer:
Calling this column `time` was ill-defined and wasn't really used for anything. Caused me some confusion when working on multiple data streams.


dylanhmorris (approved):
Thanks, @damonbayer!

PR #279, opened 2025-01-07
Point at prod_priors.py in end-to-end test

dylanhmorris:
this means the `test_output` directory can be deleted, as it only contained `priors.py`.


damonbayer (approved):
Approved, assuming the tests pass with the priors file deleted.

PR #290, opened 2025-01-09
(Do not merge) branch for test running of pyrenew H

dylanhmorris:


dylanhmorris:
It is now possible to run Pyrenew-H from #326, so this branch is no longer needed.

Issue #275, opened 2025-01-06
VARIMA baseline model for joint forecasting of NSSP and NHSN signals


SamuelBrand1:
As discussed f2f it is useful to have baseline models to compare `pyrenew` out-of-sample forecast skill.

As `pyrenew-he` is in active dev @damonbayer @dylanhmorris, I'm proposing constructing a vector ARIMA (VARIMA) model to run on NHSN jointly with ED visits. I think there are a two configurations that are worth trying:

1. For target pathogen (Flu/Covid/~RSV~) we joint forecast on NHSN and ED visits for that target.
2. For target pathogen (Flu/Covid/~RSV~) we joint forecast on NHSN and ED visits for _all_ targets. The hypothesis here would be that autocorrelations with other targets might aid forecast skill.

I'm proposing to follow other time series baselines and use `fable` to construct the model and perform parameter optimisation.

damonbayer:
Closing based on https://github.com/CDCgov/pyrenew-hew/pull/291 having been largely completed, but not merged.

PR #291, opened 2025-01-09
Issue 275: Prototype VARIMA (Analysis - do not merge)

SamuelBrand1:
This draft PR explores the VARIMA baseline model as per #275 and commits an analysis of different VARIMA structures across a few chosen dates with a few chosen ARIMA models aimed at covid and flu NHSN.

This is committed as an analysis [here](https://github.com/CDCgov/pyrenew-hew/blob/275-varima-baseline-model/demos/varima/prototype-varima.md). 

Whilst this is not an exhaustive analysis, it shows that the "best" model, as in what time series joint modelling configs have the best out-of-sample CRPS, varies over time and pathogen. This suggests that for multi-signal approaches we probably want to go towards some kind of ensemble over structures to have more consistent forecast performance.

From the pov of a baseline against pyrenew; SARIMA still seems reasonable because it was never the worst time series model.

Linking @seabbs because he was interested in VARIMA.


damonbayer:
@SamuelBrand1 @dylanhmorris Should we close this PR? or accept some part of it into the `demos` folder? It is somewhat related to https://github.com/CDCgov/pyrenew-hew/issues/346, but I think that is probably better handled in a separate PR.

damonbayer:
> @SamuelBrand1 @dylanhmorris Should we close this PR? or accept some part of it into the `demos` folder? It is somewhat related to #346, but I think that is probably better handled in a separate PR.

Bumping @SamuelBrand1

PR #302, opened 2025-01-22
Refactoring container build to use CDCgov/cfa-actions/twostep-container-build

gvegayon:


damonbayer:
@gvegayon Thanks!

Are we expecting the "Build pipeline image" to run here? I see that only the "Build dependencies image" has run.
CleanShot 2025-01-22 at 09 38 53@2x

dylanhmorris:
It looks as though the new workflow only runs a single job, which should probably be renamed (as it appears to build and push both images). Once that's done, I think we should be able to change the branch protection rule and merge. Is that correct, @gvegayon? 
https://github.com/CDCgov/pyrenew-hew/actions/runs/12900835706

gvegayon:
Done, @dylanhmorris! Regarding the pending check, I have no idea what's that!

image

I checked the workflow files, and I don't see any listing "Build pipeline image". Maybe is a legacy rule/whatnot under settings?

dylanhmorris:
>I checked the workflow files, and I don't see any listing "Build pipeline image". Maybe is a legacy rule/whatnot under settings?

Yes, that's exactly what it is and what I meant by this:

> Once that's done, I think we should be able to change the branch protection rule and merge

dylanhmorris (approved):
LGTM

PR #305, opened 2025-01-22
Make right truncation offset in forecast data None rather than 0

dylanhmorris:
Should fix the final day forecast bug (See Teams [thread](https://teams.microsoft.com/l/message/19:ff818525e4164672ba73a00e5a197406@thread.tacv2/1737561105927?tenantId=9ce70869-60db-44fd-abe8-d2767077fc8f&groupId=4f34eba7-8fcb-4a26-9b34-a434ea777f0c&parentMessageId=1737561105927&teamName=CFA-Predict&channelName=Team%20-%20STF&createdTime=1737561105927).)

Also adds a test for the `PyrenewHEWData` class and its `to_forecast_data()` method in particular


pyproject.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 python = "^3.12"
-pyrenew = {git = "https://github.com/CDCgov/PyRenew"}
+pyrenew = {git = "https://github.com/cdcgov/pyrenew"}
```
sbidari, line 10: curious what is the difference between using `github.com/CDCgov` and `github.com/cdcgov` in this case

sbidari (approved):
Looks good to me. A minor question mostly for my benefit

pyproject.toml
dylanhmorris: Nothing, and indeed this could be reverted.

PR #307, opened 2025-01-23
Update `hubverse_create_table` so that it is possible to mix epiweekly and daily other ed visit forecasts by location

dylanhmorris:
- Also adds an annotation column indicating which is which


damonbayer:
Would it be better to default to epiweekly instead of daily?

damonbayer (approved):

dylanhmorris:
> Would it be better to default to epiweekly instead of daily?

We should test this retrospectively.

PR #312, opened 2025-01-27
Improvements and bug fixes for epiweekly other hubverse tables and plots

dylanhmorris:
- Fixes bug (incorrect mixing of full paths with base location names) that caused requests to use the epiweekly other ED visit forecast to get ignored
- Makes it possible to specify a list of locations for which to use the epiweekly other forecast in `postprocess_forecast_batches.py`
- Includes epiweekly other plots in plot collation
- Adds tests that would have caught the bug

### DIFF
--- a/hewr/R/to_epiweekly_quantile_table.R
+++ b/hewr/R/to_epiweekly_quantile_table.R
@@ -65,7 +65,8 @@ to_epiweekly_quantiles <- function(model_run_dir,
       value_col = ".value"
     ) |>
     dplyr::mutate(
-      "location" = !!location
+      "location" = !!location,
+      "source_file" = !!draws_file_name
     )
   message(glue::glue("Done processing {model_run_dir}"))
   return(epiweekly_quantiles)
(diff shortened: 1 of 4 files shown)


hewr/R/to_epiweekly_quantile_table.R
```diff
@@ -122,45 +123,59 @@ to_epiweekly_quantile_table <- function(model_batch_dir,
...
+      message(glue::glue(
+        "Using epiweekly non-target ED visit forecast ",
+        "for location {loc}"
+      ))
+    } else {
+      message(glue::glue(
+        "Using daily non-target ED visit ",
+        "forecast for location {loc}."
+      ))
+    }
```
damonbayer, line 139: For concision, could use `ifelse` inside `glue` or something like `data_type <- ifelse(…)` + `glue("Using {data_type} non-target…")`

damonbayer (approved):
Thanks @dylanhmorris!

hewr/R/to_epiweekly_quantile_table.R
dylanhmorris: Done in 3293817

PR #317, opened 2025-01-31
Add wastewater data to pyrenew_hew_data

sbidari:
Added `prep_ww_data.py`: extracting, cleaning, and preprocessing NWSS data.
Updated `prep_data.py`: incorporate wastewater data into the training data.
Updated `pyrenew_hew_data.py`: downstream processing of wastewater data.
Added a test in `test_pyrenew_hew_data.py`
Misc:
Renamed some variables in the wastewater model for clarity.
Standardized date handling using `datetime.date`.

To do: Update path for NWSS data


pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -59,11 +63,21 @@ def n_hospital_admissions_datapoints(self):
...
     @property
     def n_wastewater_datapoints(self):
-        return self.get_n_datapoints(
+        return self.get_n_wastewater_datapoints(
             n_datapoints=self.n_wastewater_datapoints_,
-            data_array=self.data_observed_disease_wastewater,
+            date_array=(
+                None
+                if self.wastewater_data is None
+                else self.wastewater_data["t"].to_list()
```
dylanhmorris, line 71:
```suggestion
                else self.wastewater_data["t"].to_numpy()
```

dylanhmorris: This seems like an argument for having the wastewater data passed in as individual array-valued arguments, rather than a single data frame. But I am persuadable.

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -159,6 +250,26 @@ def get_n_datapoints(
...
+            return 0
+        elif date_array is not None and n_datapoints is not None:
+            raise ValueError(
+                "Must provide at most one out of a "
+                "number of datapoints to simulate and "
+                "an array of observed wastewater "
+                "concentration data."
+            )
+        elif date_array is not None:
+            return max(date_array)
```
dylanhmorris, line 269: What we want here is the count of days for which to sample wastewater, relative to the first data date, rather than the numboer of observations. Which suggests that this and the other two properties are badly named. Perhaps we should go with `n_wastewater_data_days`; `n_ed_visits_data_days`, etc.

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,437 @@
...
+        {
+            "date": pl.date_range(
+                start=start_date, end=end_date, interval="1d", eager=True
+            )
+        }
+    )
+    date_time_spine = date_time_spine.with_columns(
+        pl.arange(0, pl.len()).alias("t")
+    )
+    return date_time_spine
```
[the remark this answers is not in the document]
sbidari: This is yet to be moved to pyrenew-hew-data but yes after that!

pyrenew_hew/pyrenew_hew_data.py
sbidari: Yup, this is what I computed here but named the properties to be consistent with the previously defined one. Will rename!

pipelines/prep_data.py
```diff
@@ -421,9 +456,22 @@ def process_and_save_state(
         "nhsn_training_dates": nhsn_training_dates,
         "nhsn_first_date_index": nhsn_first_date_index,
         "nhsn_step_size": nhsn_step_size,
-        "state_pop": state_pop,
```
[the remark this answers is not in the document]
sbidari: Yup, put it back!

pyrenew_hew/pyrenew_hew_data.py
dylanhmorris: I think the downstream stuff _could_ be accomplished with pure `np/`jnp` but if it's simpler to do with polars what I'd do is make the dataframe from the individual arrays at computation time within `PyrenewHEWData`.

sbidari: Is there a benefit of passing individual arrays and creating a dataframe within `PyrenewHEWData` instead of passing a as a `dict`?

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,437 @@
...
+            pl.when(pl.col("lod_sewage").is_null())
+            .then(conservative_lod)
+            .otherwise(pl.col("lod_sewage"))
+            .alias("lod_sewage"),
+            pl.col("sample_collect_date")
+            .str.to_date(format="%Y-%m-%d")
+            .alias("sample_collect_date"),
+        ]
+    )
+    nwss_subset_clean = (
```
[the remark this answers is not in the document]
sbidari: > Might be nice to have an explanatory comment here. My understanding is that this is handling dates with multiple observations per day by computing an arithmetic mean observation. I think we should not do this, but rather treat the date as having two distinct observations drawn from the same error distribution about the latent value. Indeed, that will help us characterize the error distribution.

Sounds like a nice idea, but probably out of scope for this PR?

> If we are going to take a mean, it should be the geometric mean of the unlogged values (i.e. the arithmetic mean of the logged values)

Arithmetic mean if exact values repeated, geometric mean if different values?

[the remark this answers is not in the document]
sbidari: Acknowledging that this (my question) is resolved via f2f conversation.

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,179 @@ def n_days_post_init(self):
...
+                aux_subpop = pl.DataFrame()
+            site_subpop_spine = (
+                pl.concat([aux_subpop, site_indices], how="vertical_relaxed")
+                .with_columns(
+                    [
+                        pl.col("site_index").cum_count().alias("subpop_index"),
+                        pl.when(pl.col("site").is_not_null())
+                        .then(
+                            pl.col("site").map_elements(
+                                lambda x: f"Site: {x}", return_dtype=str
```
damonbayer, line 181: Can you use something more optimized from [`polars.Series.string`](https://docs.pola.rs/api/python/stable/reference/series/string.html)?

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,179 @@ def n_days_post_init(self):
...
+                    "date": pl.date_range(
+                        start=self.first_wastewater_date,
+                        end=self.last_wastewater_date,
+                        interval="1d",
+                        eager=True,
+                    )
+                }
+            )
+            return date_time_spine.with_columns(
+                pl.arange(0, pl.len()).alias("t")
```
damonbayer, line 216:
```suggestion
                pl.int_range(pl.len()).alias("t")
```

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,437 @@
...
+            )
+            | (pl.col("quality_flag").is_null())
+        )
+    )
+    conservative_lod = (
+        nwss_subset_raw.select(pl.col("lod_sewage").drop_nulls())
+        .quantile(0.95)
+        .to_numpy()
+        .item()
+    )
```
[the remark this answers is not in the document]
sbidari: what if there is no reported LOD for a lab/site pair?

pyrenew_hew/pyrenew_hew_data.py
damonbayer: better yet
```suggestion
                .with_row_index("t")
```

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,179 @@ def n_days_post_init(self):
...
+                )
+                .join(
+                    self.site_subpop_spine,
+                    on=["site_index", "site"],
+                    how="left",
+                    coalesce=True,
+                )
+                .with_columns(
+                    pl.arange(0, pl.len()).alias("ind_rel_to_observed_times")
+                )
```
damonbayer, line 234:
```suggestion
                .with_row_index("ind_rel_to_observed_times")
```

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,179 @@ def n_days_post_init(self):
...
+            )
+
+    @property
+    def pop_fraction(self):
+        if self.data_observed_disease_wastewater is not None:
+            subpop_sizes = self.site_subpop_spine.sort(
+                by="subpop_index", descending=False
+            )["subpop_pop"].to_numpy()
```
damonbayer, line 242: Sorting should not occur within this function, in my opinion. It should return the popsizes in the order that the populations are in the dataframe.

pipelines/prep_ww_data.py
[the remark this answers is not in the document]
dylanhmorris: as in use the minimum reported value for that lab-site in the `pcr_target_avg_conc` column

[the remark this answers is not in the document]
dylanhmorris: (Unless facilities routinely report values below the nominal LODs. This is worth checking in the raw data).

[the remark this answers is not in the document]
sbidari: For the particular raw data I have, about 11% of reported concentration is below LOD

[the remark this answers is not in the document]
sbidari: Resolved by dropping data for lab_sites with no reported LOD for now!

pipelines/prep_data.py
```diff
@@ -448,7 +461,21 @@ def process_and_save_state(
...
+                "site",
+                "lab",
+                "site_pop",
+                "site_index",
+                "lab_site_index",
+                "log_genome_copies_per_ml",
+                "log_lod",
+                "below_lod",
+            ]
+        ),
```
damonbayer, line 476: Any reason for this not to be what is returned by `preprocess_ww_data` so we don't have to manipulate it afterword?

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,324 @@
...
+                "location",
+            ]
+        )
+    )
+    return ww_data
+
+
+def check_missing_values(df: pl.DataFrame, columns: list[str]):
+    """Raises an error if missing values in a given column(s)."""
+    missing_cols = [col for col in columns if df[col].is_null().any()]
```
damonbayer, line 159:
```suggestion
    missing_cols = [col for col in columns if df[col].has_nulls()]
```

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,324 @@
...
+        ww_data,
+        conc_col_name=conc_col_name,
+        lod_col_name=lod_col_name,
+        date_col_name=date_col_name,
+    )
+    ww_data_ordered = ww_data.sort(by=wwtp_pop_name, descending=True)
+    lab_site_df = (
+        ww_data_ordered.select([lab_col_name, wwtp_col_name])
+        .unique()
+        .with_columns(pl.arange(0, pl.len()).alias("lab_site_index"))
```
damonbayer, line 271: another opportunity for `with_row_index`?

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,324 @@
...
+    )
+    site_df = (
+        ww_data_ordered.select([wwtp_col_name])
+        .unique()
+        .with_columns(pl.arange(0, pl.len()).alias("site_index"))
```
damonbayer, line 276: another opportunity for `with_row_index`?

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,324 @@
...
+        .with_columns(
+            [
+                (
+                    "Site: "
+                    + pl.col(wwtp_col_name).cast(pl.Utf8)
+                    + ", Lab: "
+                    + pl.col(lab_col_name).cast(pl.Utf8)
+                ).alias("lab_site_name"),
+                (pl.col("log_genome_copies_per_ml") <= pl.col("log_lod"))
+                .cast(pl.Int8)
```
damonbayer, line 298: Why?

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,324 @@
...
+    schema_overrides = {
+        "county_names": pl.Utf8,
+        "major_lab_method": pl.Utf8,
+    }
+    nwss_data = pl.read_csv(
+        ww_data_path,
+        schema_overrides=schema_overrides,
+    )  # placeholder: TBD: If using a direct API call to decipher or ABS vintage
+    ww_data = clean_and_filter_nwss_data(nwss_data).filter(
+        (pl.col("location").is_in([state_abb]))
```
damonbayer, line 320: If `state_abb` is a string, why not check for equality rather than making a singleton list?

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,324 @@
...
+        lod_col_name,
+        [
+            pl.Float32,
+            pl.Float64,
+        ],
+    )
+    check_column_type(ww_data, date_col_name, [pl.Date])
+    check_column_type(ww_data, wwtp_col_name, [pl.Int32, pl.Int64, pl.Utf8])
+    check_column_type(ww_data, lab_col_name, [pl.Int32, pl.Int64, pl.Utf8])
+    check_column_type(ww_data, wwtp_pop_name, [pl.Int32, pl.Int64])
```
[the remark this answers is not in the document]
dylanhmorris: Agree.

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,168 @@ def n_days_post_init(self):
...
+            # Check if auxiliary subpopulation needs to be added
+            add_auxiliary_subpop = (
+                self.population_size
+                > self.data_observed_disease_wastewater.select(
+                    pl.col("site_pop", "site", "lab", "lab_site_index")
+                )
+                .unique()
+                .sum()
+                .get_column("site_pop")
+                .item()
```
damonbayer, line 148: I think better to extract the column and sum it, rather than sum all the other columns (which has unclear meaning and is unnecessary).

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,168 @@ def n_days_post_init(self):
...
+                aux_subpop = pl.DataFrame(
+                    {
+                        "site_index": [None],
+                        "site": [None],
+                        "site_pop": [
+                            self.population_size
+                            - site_indices.select(pl.col("site_pop"))
+                            .sum()
+                            .get_column("site_pop")
+                            .item()
```
damonbayer, line 167: I think better to extract the column and sum it, rather than sum all the other columns (which has unclear meaning and is unnecessary).

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,168 @@ def n_days_post_init(self):
...
+                {
+                    "date": pl.date_range(
+                        start=self.first_wastewater_date,
+                        end=self.last_wastewater_date,
+                        interval="1d",
+                        eager=True,
+                    )
+                }
+            )
+            return date_time_spine.with_row_index("t")
```
damonbayer, line 210:
```suggestion
            ).with_row_index("t")
            return date_time_spine
```

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -143,7 +313,7 @@ def get_end_date(
             )
         return result
 
-    def get_n_datapoints(
+    def get_n_data_days(
```
damonbayer, line 316: Why? Should we change `n_datapoints` to `n_days` elsewhere?

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -159,16 +329,33 @@ def get_n_datapoints(
...
+        self, n_datapoints: int = None, date_array: ArrayLike = None
+    ) -> int:
+        if n_datapoints is None and date_array is None:
+            return 0
+        elif date_array is not None and n_datapoints is not None:
+            raise ValueError(
+                "Must provide at most one out of a "
+                "number of datapoints to simulate and "
+                "an array of observed wastewater "
+                "concentration data."
```
damonbayer, line 342: "an array of observed wastewater concentration data" the correct description does not really match the variable name `date_array`. Can you clarify?

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -159,16 +329,33 @@ def get_n_datapoints(
...
+            )
+        elif date_array is not None:
+            return (max(date_array) - min(date_array)).days
```
damonbayer, line 345: Seems to assume this array is daily. Is that guaranteed?

damonbayer (changes requested):
Thanks @sbidari! I left many comments. After you respond, I would like to review the whole PR again, as it was a bit much to do at one time.

pipelines/prep_ww_data.py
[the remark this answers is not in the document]
sbidari: Do you want to cast the `wwtp_col_name` and `lab_col_name` columns explicitly as strings? These are integer valued in the raw ww data.

pipelines/prep_ww_data.py
sbidari: Because I followed the convention as it was written in R, but I don't see why it can't just be a boolean datatype.

pipelines/prep_ww_data.py
sbidari: changed it!

pyrenew_hew/pyrenew_hew_data.py
sbidari: done

pyrenew_hew/pyrenew_hew_data.py
sbidari: done

pyrenew_hew/pyrenew_hew_data.py
sbidari: should be "an array of dates wastewater data is observed."

changed in the code.

pyrenew_hew/pyrenew_hew_data.py
sbidari: Why would this array have to be daily?

`date_array` contains date when wastewater data is observed at any lab/site combination. could have missing days or multiple entries on a same day

pyrenew_hew/pyrenew_hew_data.py
sbidari: https://github.com/CDCgov/pyrenew-hew/pull/317#discussion_r1945204985

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -118,12 +132,168 @@ def n_days_post_init(self):
             self.last_data_date_overall - self.first_data_date_overall
         ).days
 
+    @property
+    def site_subpop_spine(self):
```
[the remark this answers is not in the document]
sbidari: Not in this repo

pyrenew_hew/pyrenew_hew_data.py
damonbayer: I guess it depends on the expected output of `get_n_wastewater_data_days`. This will now return the number of days between the first and the last day. It does not count the number of days in the array. So, if you have 3 weeks of weekly data, it will return 21, not 3. Is that desired?

sbidari: Yes!

pipelines/prep_ww_data.py
sbidari: removed in https://github.com/CDCgov/pyrenew-hew/pull/317/commits/6cb725e38e4f268df83493bcb21a7914c623ad38

pipelines/build_pyrenew_model.py
```diff
@@ -78,6 +79,21 @@ def build_model_from_dir(
...
+        schema={
+            "date": pl.Date,
+            "site": pl.String,
+            "lab": pl.String,
+            "site_pop": pl.Int64,
+            "site_index": pl.Int64,
+            "lab_site_index": pl.Int64,
+            "log_genomes_copies_per_ml": pl.Float64,
+            "log_lod": pl.Float64,
+            "below_lod": pl.Int64,
```
damonbayer, line 93: I think this should be a boolean

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,311 @@
...
+    Creates indices for wastewater-treatment plant names and
+    flag concentration data below the level of detection.
+
+    """
+    assert (
+        conc_col_name in ww_data.columns
+    ), f"Column '{conc_col_name}' is missing in the input data."
+    assert (
+        lod_col_name in ww_data.columns
+    ), f"Column '{lod_col_name}' is missing in the input data."
```
damonbayer, line 236: Should these checks should be part of `validate_ww_conc_data`? Why only check for these columns?

pipelines/prep_ww_data.py
```diff
@@ -0,0 +1,311 @@
...
+                "log_lod",
+                "below_lod",
+            ]
+        )
+    )
+    return ww_preprocessed
+
+
+def get_nwss_data(
+    ww_data_path,
```
damonbayer, line 295: specify type for `ww_data_path`?

damonbayer (approved):
Thanks @sbidari! Please address a few remaining comments.

dylanhmorris:
Merging in the CI fix so we can check that the pipeline runs.

pipelines/prep_ww_data.py
sbidari: Good catch, I don't think we need these checks here since we explicitly create these (and other) columns in the previous step `clean_and_filter_nwss_data`

pipelines/prep_ww_data.py
sbidari: done in https://github.com/CDCgov/pyrenew-hew/pull/317/commits/6582005fb0744ea906d72ea5240142927669f4cf

pipelines/prep_ww_data.py
sbidari: Removed these and added check for all required columns in `validate_ww_conc_data` in https://github.com/CDCgov/pyrenew-hew/pull/317/commits/6582005fb0744ea906d72ea5240142927669f4cf

pipelines/build_pyrenew_model.py
sbidari: done in https://github.com/CDCgov/pyrenew-hew/pull/317/commits/6582005fb0744ea906d72ea5240142927669f4cf

PR #322, opened 2025-02-05
Simplify container build and push action

dylanhmorris:
- Removes redundant steps that are performed within https://github.com/CDCgov/cfa-actions/tree/main/twostep-container-build
- Streamlines handling of configuration variables and secrets


damonbayer (approved):
Approved, assuming you have run it and verified the metadata file we generate looks as expected.

dylanhmorris:
All looks as expected here and in the container registry

PR #326, opened 2025-02-05
Make which quantities are sampled more readily configurable

dylanhmorris:
This PR creates a clearer interface for configuring what quantities are sampled during fitting and posterior prediction. This will make it easier to use the same infrastructure for the various Pyrenew-{H,E,W} combinations.

To do:
- [x] Unit tests to confirm that configuration of what is sampled works as expected
- [x] Fix end to end test
- [x] Test fit/forecast flag combinations.


damonbayer (commented):
What do you think about an approach that parsed the model name? We already use `model_name` as an argument to many functions in the pipeline.

damonbayer:
> What do you think about an approach that parsed the model name? We already use `model_name` as an argument to many functions in the pipeline.

Per F2F discussion, the model name approach will not be taken in this PR. In the future, we could implement is and make use of the switches approach implemented in this PR.

pipelines/tests/test_end_to_end.sh
```diff
@@ -56,6 +62,7 @@ else
 	echo "TEST-MODE: Finished creating more model test data"
 fi
 
+
```
dylanhmorris, line 65:
```suggestion
```

pipelines/build_pyrenew_model.py
```diff
@@ -18,6 +18,9 @@
 
 def build_model_from_dir(
     model_dir: Path,
+    sample_ed_visits: bool = False,
+    sample_hospital_admissions: bool = False,
+    sample_wastewater: bool = False,
```
damonbayer, line 23: I don't love the word "sample" for this. Why not fit?

pipelines/generate_predictive.py
```diff
@@ -3,19 +3,30 @@
...
 
 
 def generate_and_save_predictions(
-    model_run_dir: str | Path, model_name: str, n_forecast_points: int
+    model_run_dir: str | Path,
+    model_name: str,
+    n_forecast_points: int,
+    predict_ed_visits: bool = False,
+    predict_hospital_admissions: bool = False,
+    predict_wastewater: bool = False,
```
damonbayer, line 18: Why "predict" instead of "forecast"?

pipelines/build_pyrenew_model.py
damonbayer: I guess because it gets used when building the model for forecasting.

dylanhmorris: Yes, because this is what it's actually called in the underlying pyrenew/numpyro model (which doesn't know whether it's being used for fitting, prior prediction, posterior prediction, etc.

pipelines/generate_predictive.py
dylanhmorris: Because forecasts aren't the only kind of posterior prediction this generates. Minor though.

pyrenew_hew/util.py
```diff
@@ -0,0 +1,75 @@
...
+            The relevant HEW letters, or 'null',
+    a"""
+    result = (
+        f"{'h' if fit_hospital_admissions else ''}"
+        f"{'e' if fit_ed_visits else ''}"
+        f"{'w' if fit_wastewater else ''}"
+    )
+    if result == "":
+        result = "null"
+    return result
```
damonbayer, line 40: Maybe not better:
```suggestion
hew_letters = "".join(
        char
        for cond, char in [
            (fit_hospital_admissions, "h"),
            (fit_ed_visits, "e"),
            (fit_wastewater, "w"),
        ]
        if cond
    ) or "null"
    return hew_letters
```

pipelines/batch/setup_prod_job.py
```diff
@@ -155,23 +180,41 @@ def main(
         ],
     )
 
+    needed_hew_flags = [
+        "fit_ed_visits",
+        "fit_hospital_admissions",
+        "fit_wastewater",
+        "forecast_ed_visits",
+        "forecast_hospital_admissions",
+        "forecast_wastewater",
```
damonbayer, line 189: Does the code pipeline actually work with arbitrary combinations? If not, we should check that the combination is legitimate.

dylanhmorris: I have tested all of $P(\{h, e\})$ for fitting. I have not tested all the forecast switches.

pyrenew_hew/util.py
```diff
@@ -0,0 +1,75 @@
...
+        -------
+        str
+            The relevant HEW letters, or 'null',
+    a"""
+    result = (
+        f"{'h' if fit_hospital_admissions else ''}"
+        f"{'e' if fit_ed_visits else ''}"
+        f"{'w' if fit_wastewater else ''}"
+    )
+    if result:
```
damonbayer, line 38:
```suggestion
    if not result:
```

pyrenew_hew/util.py
```diff
@@ -0,0 +1,75 @@
...
+        -------
+        str
+            The relevant HEW letters, or 'null',
+    a"""
+    result = (
+        f"{'h' if fit_hospital_admissions else ''}"
+        f"{'e' if fit_ed_visits else ''}"
+        f"{'w' if fit_wastewater else ''}"
+    )
+    if result == "":
```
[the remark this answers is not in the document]
damonbayer: Oops https://github.com/CDCgov/pyrenew-hew/pull/326#discussion_r1951464148

pipelines/batch/setup_prod_job.py
dylanhmorris: Testing complete. Forecasting more signals than you fit is currently supported. Not forecasting stuff you _do_ fit breaks the `forecast_state.py` pipeline as currently configured.

dylanhmorris: Added some temporary errors to `setup_prod_job` and `forecast_state.py` to prevent attempting this. The issue is with the postprocessing rather than with the fitting (or posterior prediction) itself. We should discuss whether we want to permit this or not. I think it is low priority because in general we will want to forecast things we're fitting to when running `forecast_state.py` (given the name!)

damonbayer (approved):
Thanks @dylanhmorris!

Issue #321, opened 2025-02-04
Intermittent failures on Docker push to ACR


dylanhmorris:
Typically resolved by re-running. See e.g. https://github.com/CDCgov/pyrenew-hew/actions/runs/13122404681/job/36670092161 (run triggered by push to `main` failed; manual re-run succeeded)

See also #314

dylanhmorris:
@gvegayon any insight?

gvegayon:
Hey @dylanhmorris, I saw something similar in the past. I can't find the reference, but it is an error others have seen pushing/pulling images from Azure CR. In some cases, I recall, it had to do with the size of the image, but what you have here is not that big, last time I checked.

dylanhmorris:
Update, it seems very, very likely that it is this: https://github.com/docker/login-action/issues/173

We see it most frequently on merges to main. This makes sense because those merges to main also trigger the delete container tag action. This begins after the container build action but finishes first (and logs out).

https://github.com/CDCgov/pyrenew-hew/actions/runs/13190112728

gvegayon:
This can be closed now, right, @dylanhmorris?

dylanhmorris:
Not quite. We haven't yet dealt with the potential conflict if two build/push actions run simultaneously 

> Docker login not needed since we log in directly to Azure and use az. This will also make https://github.com/CDCgov/pyrenew-hew/issues/321 rarer (though there we might still have problems with two build/push workflows running simultaneously)

[Link](https://github.com/CDCgov/pyrenew-hew/pull/329)

gvegayon:
Referencing [this issue](https://github.com/CDCgov/cfa-actions/issues/12) that may be related.

dylanhmorris:
As of #341 we do not use the docker login action anywhere in this repo, so this can be closed.

PR #329, opened 2025-02-07
Streamline tag delete

dylanhmorris:
Docker login not needed since we log in directly to Azure and use `az`. This will also make #321 rarer (though there we might still have problems with two build/push workflows running simultaneously)


damonbayer (approved):

gvegayon:
I'm reporting this on cfa-epinow2-pipeline. Thanks for catching this, Dylan!

PR #334, opened 2025-02-10
Specify explicit path for pipelines and pyrenew-hew tests

### LINKED ISSUE #333: An error in either `pyrenew-hew/tests` or `pipelines/tests` causes both jobs to fail
I think this is on me and specifying an explicit path while running the test should fix it

sbidari:
fixes #333


damonbayer (approved):
Good catch

PR #341, opened 2025-02-11
Use SP authentication in containers.yaml

dylanhmorris:
Instead of a hard-coded account password.

### DIFF
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -27,7 +27,7 @@ jobs:
           creds: ${{ secrets.EDAV_STF_SERVICE_PRINCIPAL }}
 
       - name: Login to ACR
-        run: az acr login --name ${{ vars.CONTAINER_REGISTRY }}
+        run: az acr login --name ${{ vars.CONTAINER_REGISTRY_NAME }}
         
       - name: Build and push image
         id: build-push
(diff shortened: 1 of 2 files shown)


.github/workflows/containers.yaml
```diff
@@ -20,16 +20,22 @@ jobs:
     name: Build and push image
 
     steps:
+      - name: Login to Azure
+        id: azure_login_2
+        uses: azure/login@v2
+        with:
+          creds: ${{ secrets.EDAV_STF_SERVICE_PRINCIPAL }}
```
damonbayer, line 27: Is there documentation for this usage? It's different than our solution in pipeline check action? https://github.com/CDCgov/pyrenew-hew/blob/main/.github/workflows/pipeline-run-check.yaml#L19-L21

dylanhmorris: We don't need the secret as an environment variable here; we can just pass it directly to the azure login action. Conversely, in `pipeline-run-check`, we need the API username and secret as environment variables because that's where `forecasttools::pull_nhsn` looks for them.

dylanhmorris: Note that this is the approach taken in: https://github.com/CDCgov/pyrenew-hew/blob/main/.github/workflows/delete-container-tag.yaml

I am currently proposing to add it to the docs of `twostep-container-build`: https://github.com/CDCgov/cfa-actions/pull/14

damonbayer (approved):

PR #353, opened 2025-02-14
update end to end test

sbidari:
Key changes:

Adds `generate_fake_nwss_data` function in `generate_test_data.R` 
Modifies `end_to_end_test.sh` to fit `Pyrenew-HEW` model for COVID-19 

Adds `PyrenewWastewaterData` class to handle wastewater data.
Modifies `build_model_from_dir` and related functions to incorporate wastewater data.
Adds priors for wastewater-related parameters in `prod_priors.py` and `build_pyrenew_model.py`
Modifies `forecast_state.py` to handle wastewater data:
-scan for NWSS data conditional on `fit_wastewater` flag
-raise errors if wastewater data is not available for the requested report date.
Moved calculation of `pop_fraction` from `PyrenewHEWData` to `prep_data.py` (required arg for `LatentInfectionProcess`)


pyrenew_hew/pyrenew_wastewater_data.py
```diff
@@ -0,0 +1,177 @@
...
+            if add_auxiliary_subpop:
+                aux_subpop = pl.DataFrame(
+                    {
+                        "site_index": [None],
+                        "site": [None],
+                        "site_pop": [
+                            self.population_size
+                            - site_indices.select(pl.col("site_pop"))
+                            .get_column("site_pop")
+                            .sum()
```
dylanhmorris, line 57:
```suggestion
                            - site_indices.get_column("site_pop").sum()
```

pyrenew_hew/pyrenew_wastewater_data.py
```diff
@@ -0,0 +1,177 @@
...
+                        ],
```
dylanhmorris, line 58:
```suggestion
                        "site_pop": list(
                            self.population_size - site_indices.get_column("site_pop").sum()
                        ),
```

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -19,19 +20,22 @@ def __init__(
...
-        shedding_offset: float = 1e-8,
-        pop_fraction: ArrayLike = jnp.array([1]),
+        wastewater_data: PyrenewWastewaterData = None,
+        n_ww_lab_sites: int = None,
+        ww_censored: ArrayLike = None,
+        ww_uncensored: ArrayLike = None,
+        ww_observed_subpops: ArrayLike = None,
+        ww_observed_times: ArrayLike = None,
+        ww_observed_lab_sites: ArrayLike = None,
+        lab_site_to_subpop_map: ArrayLike = None,
```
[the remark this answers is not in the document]
sbidari: I need to pass `data_observed_disease_wastewater` to `PyrenewWastewaterData` to get these (`n_ww_lab_sites` and other attributes) but this also makes `data_observed_disease_wastewater_conc` not `None` when calling `to_forecast_data()`

sbidari:
@dylanhmorris I have made the changes we discussed yesterday.

This passes the pipeline run check but fails on a test for `PyrenewWastewaterData`. The issue is that `site_pop` (within `site_subpop_spine`) is interpreted as Object when accessed from `build_model_from_dir` but is interpreted as Float when called from `test_pyrenew_wastewater_data` leading to this error. I don't really understand what's causing this difference

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -120,9 +121,9 @@ def sample(self, n_days_post_init: int):
             log_rtu_weekly_subpop = log_rtu_weekly[:, jnp.newaxis]
         else:
             i_first_obs_over_n_ref_subpop = transformation.SigmoidTransform()(
-                transformation.logit(i0_first_obs_n)
+                transforms.logit(i0_first_obs_n)
```
dylanhmorris, line 124: Can use `transformation.SigmoidTransform().inv()`

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -136,7 +137,7 @@ def sample(self, n_days_post_init: int):
                 DistributionalVariable(
                     "i_first_obs_over_n_non_ref_subpop_raw",
                     dist.Normal(
-                        transformation.logit(i0_first_obs_n),
+                        transforms.logit(i0_first_obs_n),
```
dylanhmorris, line 140:
```suggestion
                        transformation.SigmoidTransform().inv(i0_first_obs_n),
```

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -5,6 +5,7 @@
 import jax.numpy as jnp
 import numpyro
 import numpyro.distributions as dist
+import numpyro.distributions.transforms as transforms
```
dylanhmorris, line 8: Can remove if you take my suggestions below.
```suggestion
```

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -354,6 +212,13 @@ def to_forecast_data(self, n_forecast_points: int) -> Self:
             first_ed_visits_date=self.first_data_date_overall,
             first_hospital_admissions_date=(self.first_data_date_overall),
             first_wastewater_date=self.first_data_date_overall,
-            pop_fraction=self.pop_fraction,
```
dylanhmorris, line 357: do we not still need this?

pyrenew_hew/pyrenew_wastewater_data.py
```diff
@@ -0,0 +1,202 @@
...
+    def to_pyrenew_hew_data_args(self):
+        return {
+            attr: value
+            for attr, value in (
+                (attr, getattr(self, attr))
+                for attr, prop in self.__class__.__dict__.items()
+                if isinstance(prop, property)
+            )
+            if isinstance(value, ArrayLike)
+        }
```
dylanhmorris, line 202: I think it's more readable just to enumerate here. There aren't that many args. I think something like this is fine:

```python
return {
   "a": self.a,
   "b:  self.b
}
```
or if you want to be a bit less verbose:

```python
return { arg: getattr(self, arg) for attr in ["list", "the", "args", "here"]} 
```

dylanhmorris (commented):
One question and one suggestion. Thanks, @sbidari!

pyrenew_hew/pyrenew_hew_data.py
sbidari: I believe `pop_fraction` needs to be passed to `forecast_data`. Surprisingly, the model runs to completion even when `forecast_data.pop_fraction` is `None`

pyrenew_hew/pyrenew_wastewater_data.py
sbidari: changed to use second suggestion in https://github.com/CDCgov/pyrenew-hew/pull/353/commits/8899552ab05d335a2ed7f3fd4ef8f06316670c07

pyrenew_hew/pyrenew_hew_data.py
sbidari: added `pop_fraction` and `ww_log_lod` here. Test run locally threw an error for missing `ww_log_lod` but no error if `forecast_data.pop_fraction = None`. Will investigate separately if we want to review and merge this PR

dylanhmorris (approved):
LGTM provided tests pass, thanks @sbidari!. In the future we should more fully separate the subpopulation specification from the wastewater data specification (so you can do multi-subpop models without observed wastewater). This should also help with the pop-fraction stuff. But it's out of scope for now.

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -120,9 +120,8 @@ def sample(self, n_days_post_init: int):
             log_rtu_weekly_subpop = log_rtu_weekly[:, jnp.newaxis]
         else:
             i_first_obs_over_n_ref_subpop = transformation.SigmoidTransform()(
-                transformation.logit(i0_first_obs_n)
+                transformation.SigmoidTransform().inv(i0_first_obs_n)
                 + self.offset_ref_logit_i_first_obs_rv(),
```
dylanhmorris, line 124:
```suggestion
                + self.offset_ref_logit_i_first_obs_rv(),
            )
```

PR #354, opened 2025-02-18
Infer offset between infection to ED visit delay distribution and infection to hosp admission delay distribution

dylanhmorris:


hewr/R/to_epiweekly_quantile_table.R
```diff
@@ -101,6 +101,7 @@ to_epiweekly_quantile_table <- function(model_batch_dir) {
       )
 
       forecast_data |>
+        dplyr::filter(.data$.variable %in% names(variable_target_key)) |>
```
damonbayer, line 104: This is apparently necessary, based on some production problems we had without it, but I don't immediately see why.

dylanhmorris: Wastewater variables for HW and HEW I suspect.

dylanhmorris: which don't currently have a hubverse table name in the `variable_target_key` key.

damonbayer: But we are already running W models, so why do they need to be dealt with differently than the existing solution in this PR?

damonbayer: I actually ended up having to implement this in https://github.com/CDCgov/pyrenew-hew/pull/388, but I'm not sure why it isn't causing issues in `main` right now.

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -27,6 +28,59 @@
 from pyrenew_hew.pyrenew_hew_data import PyrenewHEWData
 
 
+class OffsetDiscretizedLognormalPMF(RandomVariable):
```
sbidari, line 31: Why not move this to `PyRenew` repo?

dylanhmorris:
This is now ready.

pipelines/score_forecast.R
damonbayer, line 1: Just wondering why this file received changes. Are you using Air locally?

dylanhmorris: Not yet in this repo, so idk.

dylanhmorris: different versions of `styler`?

dylanhmorris: I think I see what it is. I think it's left over from a manual conflict resolution when I merged your changes from #388

dylanhmorris: Incidentally, that comment is now out of date, no?

damonbayer: I still see that the file is modified compared to `main`

pipelines/prep_data.py
```diff
@@ -365,9 +371,17 @@ def get_pmfs(param_estimates: pl.LazyFrame, state_abb: str, disease: str):
...
         .collect(streaming=True)
         .get_column("value")
-        .to_list()[0]
+        .item(0)
+        .to_list()
     )
 
+    # ensure 0 first entry; we do not model the possibility
+    # of a zero infection-to-recorded-admission delay in Pyrenew-HEW
+    delay_pmf[0] = 0.0
```
damonbayer, line 380: What is `delay_pmf[0]` before re-assignment? (Just want to check we shouldn't instead be prepending the `0.0`)

pipelines/score_forecast.R
dylanhmorris: I meant the code comment in the file:

```
  # Removing lab_site_index column
  # NWSS data is not added to eval data yet
```

pipelines/prep_data.py
```diff
@@ -378,12 +392,73 @@ def get_pmfs(param_estimates: pl.LazyFrame, state_abb: str, disease: str):
...
         .get_column("value")
-        .to_list()[0]
+        .item(0)
+        .to_list()
     )
 
     return (generation_interval_pmf, delay_pmf, right_truncation_pmf)
 
 
+def approx_lognorm(
```
damonbayer, line 402: Seems like it would be easy enough to do analytically, but probably not important enough to change.

pipelines/prep_data.py
```diff
@@ -418,6 +493,12 @@ def process_and_save_state(
         param_estimates=param_estimates, state_abb=state_abb, disease=disease
     )
 
+    delay_lognormal_loc, delay_lognormal_scale = approx_lognorm(
```
damonbayer, line 496: refer to `inf_to_hosp_admit` in the name for clarity?

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -27,6 +28,59 @@
...
+    scale of a lognormal distribution from central
+    values, then discretizing and normalizing.
+    """
+
+    def __init__(
+        self,
+        name,
+        reference_loc: ArrayLike,
+        reference_scale: ArrayLike,
+        n: int,
```
damonbayer, line 43: document arguments? Or choose a more descriptive name than `n`.

pipelines/priors/prod_priors.py
```diff
@@ -89,9 +96,9 @@
     "ed_visit_neg_bin_concentration", dist.LogNormal(4, 1)
 )
 
-# more diffuse than ED visit, same mean
+# should be approx 7-fold more concentrated, all else equal
```
damonbayer, line 99: Unclear to me what this means or if the code accomplishes what you intended.

Playing around with some actual numbers:
```python
import jax.numpy as jnp
from numpyro import distributions as dist

original_conc = dist.LogNormal(4, 1.5)

sevenfold_conc = dist.LogNormal(4 + jnp.log(7), 1.5)

dist.NegativeBinomial2(1000.0, original_conc.mean).variance

dist.NegativeBinomial2(1000.0, sevenfold_conc.mean).variance
```

dylanhmorris: It's not about the variance, it's about the actual concentration parameter ($k$, $\phi$, $\theta$). The point here is that the sum of $n$ neg bin RVs with mean $\mu$ and conc $k$ is neg bin with mean $n \mu$ and conc $n k$.

pipelines/score_forecast.R
damonbayer: I think that's right.

damonbayer (approved):
Please address comments, but nothing showstopping.

pipelines/score_forecast.R
dylanhmorris: Removed in 7a207a6

pipelines/prep_data.py
dylanhmorris: It does represent a 0 delay (in the NHSN-inferred PMF) and is small.

pipelines/prep_data.py
dylanhmorris: I don't really want to stick with this approach long term, so I think this is fine.

pipelines/prep_data.py
dylanhmorris: 4f363f

pipelines/build_pyrenew_model.py
```diff
@@ -53,28 +55,43 @@ def build_model_from_dir(
...
+    # For now follow NNH in just substituting
+    # (eventually will use a different inferred fixed).
+    if he:
+        # offset from approx inf to ed distribution
+        # when fitting admissions
+        inf_to_hosp_admit_rv = OffsetDiscretizedLognormalPMF(
+            "inf_to_hosp_admit",
+            reference_loc=model_data["inf_to_hosp_admit_lognormal_loc"],
+            reference_scale=model_data["inf_to_hosp_admit_lognormal_scale"],
+            n=jnp.size(model_data["inf_to_hosp_admit_pmf"]) * 2,
```
damonbayer, line 80: Not following the reasoning for `* 2`.

dylanhmorris: Flexibility to infer delays with a longer tail, up to a point.

dylanhmorris, line 80:
```suggestion
            n=jnp.size(model_data["inf_to_hosp_admit_pmf"]) * 2,
            # Flexibility to infer delays with a longer tail, up to a point.
```

dylanhmorris: added an explanatory comment

PR #357, opened 2025-02-19
Rename plots so that `log` plots are alphabetically adjacent to non-`log` plots

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/356

### DIFF
--- a/pipelines/plot_and_save_state_forecast.R
+++ b/pipelines/plot_and_save_state_forecast.R
@@ -62,8 +62,8 @@ save_forecast_figures <- function(model_run_dir,
         pyrenew_model_name,
         glue(
           "{target_variable}_",
-          "forecast_plot{transform_name}_",
-          "{timescale}"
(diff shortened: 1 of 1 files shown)


pipelines/plot_and_save_state_forecast.R
```diff
@@ -62,8 +62,8 @@ save_forecast_figures <- function(model_run_dir,
         pyrenew_model_name,
         glue(
           "{target_variable}_",
-          "forecast_plot{transform_name}_",
-          "{timescale}"
+          "forecast_plot_{timescale}",
```
dylanhmorris, line 65:
```suggestion
          "{timescale}",
```

dylanhmorris: I think now is also a good opportunity to shorten plot names

dylanhmorris (approved):

PR #359, opened 2025-02-19
Clean up `build_pyrenew_model.py`, add absolute versus relative IHR auto-switch

dylanhmorris:
## Changes
Small PR to make variable names a bit short and more meaningful in `build_pyrenew_model.py` and configure the absolute versus relative IHR via the fitting flags.


dylanhmorris (commented):
Self-reviewing.

PR #361, opened 2025-02-27
Install `pak` and `arrow` as deps in Containerfile.dependencies

dylanhmorris:


dylanhmorris:
Not convinced this is a desirable approach, so closing this.

PR #365, opened 2025-02-27
Suggestion for how to handle the population calculation

dylanhmorris:
This fixes the failing test for me locally.

### DIFF
--- a/pyrenew_hew/pyrenew_wastewater_data.py
+++ b/pyrenew_hew/pyrenew_wastewater_data.py
@@ -43,7 +43,7 @@ class PyrenewWastewaterData:
                 .sum()
             )
 
-            total_pop_no_ww = self.population_size - total_pop_ww > 0
+            total_pop_no_ww = self.population_size - total_pop_ww
 
             add_auxiliary_subpop = total_pop_no_ww > 0
 
--- a/pyrenew_hew/pyrenew_wastewater_data.py
+++ b/pyrenew_hew/pyrenew_wastewater_data.py
@@ -43,8 +43,9 @@ class PyrenewWastewaterData:
                 .sum()
             )
 
-            total_pop_no_ww = self.population_size - total_pop_ww
-
+            total_pop_no_ww = (
+                int(jnp.atleast_1d(self.population_size)[0]) - total_pop_ww
+            )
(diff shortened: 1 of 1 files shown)


pyrenew_hew/pyrenew_wastewater_data.py
```diff
@@ -27,31 +27,32 @@ def __init__(
...
+
+            total_pop_ww = (
+                self.data_observed_disease_wastewater.unique(
+                    ["site_pop", "site"]
+                )
+                .get_column("site_pop")
+                .sum()
+            )
+
+            total_pop_no_ww = self.population_size - total_pop_ww > 0
```
dylanhmorris, line 46:
```suggestion
            total_pop_no_ww = self.population_size - total_pop_ww
```

pyrenew_hew/pyrenew_wastewater_data.py
```diff
@@ -27,31 +27,33 @@ def __init__(
...
+                self.data_observed_disease_wastewater.unique(
+                    ["site_pop", "site"]
+                )
+                .get_column("site_pop")
+                .sum()
+            )
+
+            total_pop_no_ww = (
+                int(jnp.atleast_1d(self.population_size)[0]) - total_pop_ww
+            )
```
damonbayer, line 48: I don't understand all the complication here.

1. Why do we need to use `jnp.atleast_1d`?
2. Why do we need to cast to `int`?

dylanhmorris: It being a `jnp.array` of unknown shape elsewhere, which @sbidari is now fixing.

sbidari:
After some trial and error, I found adding the following in `build_model_from_dir` resolves the issue.
```{python}
    population_size = jnp.array(model_data["state_pop"]).item()
```
 Thanks @dylanhmorris for your help with identifying the issue.
I will plan to close this PR without merging unless you'd like me to fold in the changes made to `site_subpop_spine` in this PR

dylanhmorris:
I think it would be good to fold in those changes, as I do think they make the logic a little simpler/clearer.

sbidari (approved):
Thanks @dylanhmorris

PR #371, opened 2025-03-06
H,E,W component tests

### LINKED ISSUE #336: Modularize end to end test runs
Once #326 lands, we should revise the end-to-end test take the following form

1. Fit each of Pyrenew- $\mathcal{P}(\\{H,E,W\\})$ (except for Pyrenew- $\emptyset$) for each of {US, CA, MT}, in separate subprocesses, and upload results
2. Once all are complete, download results and run postprocessing on all.

This will be faster and will better simulate the actual production workflow.

dylanhmorris:
Will close #336


pipelines/forecast_state.py
```diff
@@ -648,45 +657,28 @@ def get_available_nwss_reports(
...
+    fit_flags = flags_from_hew_letters(args.model_letters)
+    additional_forecast_flags = flags_from_hew_letters(
+        args.additional_forecast_letters
+    )
+    forecast_flags = {
+        f"forecast_{x}": (
+            fit_flags[f"fit_{x}"] or additional_forecast_flags[f"fit_{x}"]
+        )
+        for x in ["hospital_admissions", "ed_visits", "wastewater"]
+    }
```
damonbayer, line 681:
```suggestion
    forecast_flags = [
        f"forecast_{x}"
        for x in set(args.model_letters)
        | set(args.additional_forecast_letters)
    ]
```

damonbayer, line 681:
```suggestion
    forecast_flags = {
        f"forecast_{x}"
        for x in args.model_letters + args.additional_forecast_letters
    }
```

damonbayer, line 681: Seems like an additional argument to `flags_from_hew_letters` could be really useful for tidying this up.

pipelines/tests/test_end_to_end.sh
```diff
@@ -1,124 +1,53 @@
...
-    echo "Usage: $0 <base_dir>"
+BASE_DIR=pipelines/tests/end_to_end_test_output
+echo "TEST-MODE: Running forecast_state.py in test mode with base directory $BASE_DIR"
+
+if [ -d $BASE_DIR ]
+then
+    # make the user delete the directory, to avoid accidental deletes of
+    # test output
+    echo "TEST-MODE FAIL: test output directory $BASE_DIR already exists."
+    echo "DETAILS: The test output directory persists after each run to allow the user to examine output. It must be deleted and recreated at the start of each new end-to-end test run to ensure that old output does not compromise test validity. Delete the directory and re-run the test."
```
damonbayer, line 11: Command line arg to delete the directory would be useful for running locally while debugging.

pipelines/tests/test_end_to_end.sh
```diff
@@ -1,124 +1,53 @@
...
-				--fit-ed-visits \
-				--fit-hospital-admissions \
-				--no-fit-wastewater \
-				--forecast-ed-visits \
-				--forecast-hospital-admissions \
-				--no-forecast-wastewater \
-				--score \
-				--eval-data-path "$BASE_DIR/private_data/nssp-etl"
-if [ $? -ne 0 ]; then
+	    if [[ ($model == *w* && ($disease == "Influenza"  || $location == "US") ) || $model == "w" ]]
```
damonbayer, line 31: I don't fully understand why we use this filter. Can you include a comment?

pipelines/forecast_state.py
dylanhmorris: I agree and had been debating whether to do in this PR. Now done in a5cd721

pipelines/tests/test_build_pyrenew_model.py
```diff
@@ -65,13 +65,13 @@ def mock_priors():
 ihr_rel_iedr_rv = None
 t_peak_rv = None
 duration_shed_after_peak_rv = None
-inf_to_ed_offset_loc_rv = None
-inf_to_ed_log_offset_scale_rv = None
+delay_offset_loc_rv = None
+delay_log_offset_scale_rv = None
```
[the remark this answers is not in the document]
dylanhmorris: That snuck in accidentally from #354. Will fix.

.github/workflows/pipeline-run-check.yaml
damonbayer, line 1: Lots of repeated steps. Suggest creating a composite action to setup the environment and cache dependencies.

pyrenew_hew/util.py
```diff
@@ -113,11 +127,11 @@ def flags_from_hew_letters(hew_letters: str) -> dict[str, bool]:
...
         )
-    return dict(
-        fit_hospital_admissions="h" in hew_letters,
-        fit_ed_visits="e" in hew_letters,
-        fit_wastewater="w" in hew_letters,
-    )
+    return {
+        f"{flag_prefix}_hospital_admissions": "h" in hew_letters,
+        f"{flag_prefix}_ed_visits": "e" in hew_letters,
+        f"{flag_prefix}_wastewater": "w" in hew_letters,
```
damonbayer, line 133: Probably not worth the effort to change, but I question why we didn't just use `{flag_prefix}_{letter}`, instead of expanding to the full names.

damonbayer (approved):
A few small changes requested. Thanks @dylanhmorris!

pipelines/tests/test_end_to_end.sh
dylanhmorris: done in 1d51adb

pipelines/tests/test_end_to_end.sh
dylanhmorris: In lieu of a comment, updated the test skipping message that occurs conditional on the filter to add more or the motivation. Does that resolve it?

pyrenew_hew/util.py
dylanhmorris: The motivation was to spell things out for readability, but I take your point that we use enough of `H`, `E`, and `W` throughout that it's probably not an issue.

.github/workflows/pipeline-run-check.yaml
dylanhmorris: Agree but separate PR I think, especially since we'll want to incorporate it into other workflows as well.

dylanhmorris: Opened #374

pyrenew_hew/util.py
dylanhmorris: Opened #375

PR #377, opened 2025-03-06
rename generate_epiweekly to generate_epiweekly_data

damonbayer:
I am constantly confusing this with code that aggregates daily model output


dylanhmorris (approved):
LGMT, thanks!

Issue #415, opened 2025-04-07
Improve printing of error messages from Python subprocess calls


dylanhmorris:
We currently have constructs like this:

https://github.com/CDCgov/pyrenew-hew/blob/850042f05ed0727f37158cb33d06a5166783efd3/pipelines/postprocess_forecast_batches.py#L73-L83

When errors do occur, the captured `stderr` output prints as a raw string, with line breaks as `\n`, e.g.:

```bash
RuntimeError: create_hubverse_table: b'Error in `purrr::map()`:\n\xe2\x84\xb9 In index: 36.\n\xe2\x84\xb9 With name:\n  private_data/mounts/pyrenew-hew-prod-output/2025-04-07_forecasts/covid-19_r_2025-04-07_f_2025-01-06_t_2025-04-05/model_runs/OK.\nCaused by error in `purrr::map()`:\n\xe2\x84\xb9 In index: 4.\n\xe2\x84\xb9 With name:\n  private_data/mounts/pyrenew-hew-prod-output/2025-04-07_forecasts/covid-19_r_2025-04-07_f_2025-01-06_t_2025-04-05/model_runs/OK/timeseries_e/daily_baseline_ts_forecast_samples.parquet.\nCaused by error in `dplyr::reframe()`:\n\xe2\x84\xb9

<more>
```

This makes troubleshooting/debugging unpleasant, and so it is probably worth the effort to improve.

PR #416, opened 2025-04-07
Improve printing of error messages from Python subprocess calls

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/415


dylanhmorris (approved):
I have tried this out in practice and confirmed it resolves the issue. Thanks @damonbayer!

Issue #395, opened 2025-03-24
Migrate from Poetry to UV


damonbayer:
https://github.com/cdcent/cfa-predict-handbook/issues/446#issuecomment-2741758480

dylanhmorris:
@damonbayer confirming that you are planning to do this this week, yes?

damonbayer:
Yes, I'm waiting to verify https://github.com/CDCgov/pyrenew-hew/pull/400 works as expected before making changes to the images.

PR #444, opened 2025-04-24
Migrate from Poetry to UV

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/395


damonbayer:
@natemcintosh has some really excellent guidance [here](https://github.com/cdcent/cfa-nssp-etl/pull/100), but I'm struggling a bit with our more complicated docker setup.

natemcintosh:
what is going wrong with the containerfile?

Containerfile
```diff
@@ -27,8 +31,13 @@ RUN Rscript -e "pak::pkg_install('cmu-delphi/epiprocess@main')"
...
 RUN Rscript -e "pak::local_install('hewr')"
 
+
 COPY --exclude=pipelines/priors . .
-RUN pip install --root-user-action=ignore .
+
+# Python
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync
+RUN uv add quarto-cli
```
dylanhmorris, line 40: Why use `uv` rather than base `pip` for the container?

dylanhmorris: (Given that it's a static, not a dev installation)

natemcintosh: I think either works, but `uv` should be considerably faster, especially if it can use a cache like here.

damonbayer:
> what is going wrong with the containerfile?

I think everything is working fine, but I am a novice at docker, so could use a thorough review from someone more experienced to see if am doing anything sub-optimally.

Containerfile
```diff
@@ -1,21 +1,25 @@
...
-FROM python:3.13
-
-ARG PYRENEW_VERSION="v0.1.2"
-RUN apt-get update
-RUN apt-get install -y r-base
-RUN apt-get install -y cmake
-RUN pip install --root-user-action=ignore -U pip
-RUN pip install --root-user-action=ignore git+https://github.com/cdcgov/pyrenew.git@$PYRENEW_VERSION
-RUN pip install --root-user-action=ignore quarto-cli
+FROM python:3.13-slim-bookworm
```
damonbayer, line 3: Copied this from @natemcintosh. I'm not sure if some other image would be a better base.

Containerfile
damonbayer, line 40: quarto isn't really used anymore, since we have soft-deprecated the dashboard.

Containerfile
```diff
@@ -1,21 +1,25 @@
...
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
 ARG GIT_COMMIT_SHA
 ENV GIT_COMMIT_SHA=$GIT_COMMIT_SHA
 
 ARG GIT_BRANCH_NAME
 ENV GIT_BRANCH_NAME=$GIT_BRANCH_NAME
```
damonbayer, line 10: It would be nice to get rid of this and just have the container image have the full repo, so we can access various git more directly.

dylanhmorris: we've discussed this for a while and I agree.

Containerfile
natemcintosh: That's just the one the [uv docs](https://docs.astral.sh/uv/guides/integration/docker/#available-images) use. So far have had no issues with it, but it's possible there are better options depending on the use case.

Containerfile
natemcintosh: I'd go ahead and remove it if you aren't using anymore

Containerfile
```diff
@@ -1,21 +1,25 @@
 #syntax=docker/dockerfile:1.7-labs
 
-FROM python:3.13
-
-ARG PYRENEW_VERSION="v0.1.2"
-RUN apt-get update
-RUN apt-get install -y r-base
-RUN apt-get install -y cmake
-RUN pip install --root-user-action=ignore -U pip
-RUN pip install --root-user-action=ignore git+https://github.com/cdcgov/pyrenew.git@$PYRENEW_VERSION
```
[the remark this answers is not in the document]
natemcintosh: Since PyRenew is listed in your main `pyproject` dependencies, it will be installed automatically in the `uv sync` step.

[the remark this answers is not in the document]
damonbayer: I believe we were installing separately for caching-related reasons. Perhaps @dylanhmorris can comment further.

Containerfile
natemcintosh: Especially the explanation of the variants towards the bottom of the page.

dylanhmorris (approved):
LGTM. Thanks, @damonbayer!

Containerfile
dylanhmorris: #446

Containerfile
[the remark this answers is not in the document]
dylanhmorris: Yes, that's right. But this might not be a problem any longer for the reasons @natemcintosh mentions. I will investigate when I address #446

Containerfile
dylanhmorris: #445

Containerfile
dylanhmorris: Flagging to review in #446

PR #447, opened 2025-04-25
Remove stray commas that prevented `--help` from working in `setup_job.py`

damonbayer:


dylanhmorris:
Goot catch, thanks! This had been bothering me for a while but I had never investigated.

PR #456, opened 2025-04-28
updating workflow to delete image tags from ghcr

giomrella:
Using a third-party action to delete image tags by name. This action has been approved for use with cdcent and cdcgov by the CDC Github team. Using the Github REST API to delete image tags is surprisingly cumbersome.

### DIFF
--- a/.github/workflows/delete-container-tag.yaml
+++ b/.github/workflows/delete-container-tag.yaml
@@ -2,6 +2,11 @@ name: Delete branch tag from container registry
 
 on:
   delete:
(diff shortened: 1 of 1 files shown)


giomrella:
I added a workflow_dispatch trigger to this workflow so it can be used to manually clean up existing images and so I can test the action before merging, but it looks like I can't manually run the action until it has been merged into main. I have tested this action on user-owned repos with the GITHUB_TOKEN, not org repos, but it should still work.

Issue #461, opened 2025-04-28
Use ggdist 3.3.3 from CRAN


damonbayer:
We no longer need to use dev ggdist, with the release of 3.3.3.

PR #463, opened 2025-04-28
Use ggdist 3.3.3 from CRAN

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/461

### DIFF
--- a/hewr/DESCRIPTION
+++ b/hewr/DESCRIPTION
@@ -20,7 +20,7 @@ Imports:
     forcats,
     forecasttools (>= 0.1.6),
     fs,
-    ggdist,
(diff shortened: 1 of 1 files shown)


dylanhmorris:
@damonbayer can you investigate the failing R CMD check workflow? Re-run didn't resolve. Suspect clearing cache and re-running might.

damonbayer:
> @damonbayer can you investigate the failing R CMD check workflow? Re-run didn't resolve. Suspect clearing cache and re-running might.

I deleted caches repeatedly, and it eventually worked. Not sure if it has anything to do with the cache.

dylanhmorris (approved):
LGTM

PR #465, opened 2025-04-29
only r cmd check on latest r version

damonbayer:


dylanhmorris (approved):
Seems reasonable given how we use this software

PR #422, opened 2025-04-10
Make datetime handling more robust

sbidari:
See #327

PR #473, opened 2025-05-01
update observed_{signal} indices

sbidari:
This PR includes changes to #422 from the last approved changes.


sbidari:
I have run the model with real data for several states and all plots look good

damonbayer:
Figures look good @sbidari. Is there a way we could implement a test that would have caught these problems?

hewr/R/process_state_forecast.R
```diff
@@ -372,6 +373,25 @@ process_pyrenew_model <- function(model_run_dir,
...
+    dplyr::summarise(predicted_last_date = max(date)) |>
+    dplyr::mutate(
+      expected_last_date = dplyr::case_when(
+        stringr::str_ends(.variable, "ed_visits") ~
+          last_data_date_overall + n_forecast_days,
+        .variable == "site_level_log_ww_conc" ~ last_data_date_overall,
+        stringr::str_ends(.variable, "hospital_admissions") ~
+          forecasttools::ceiling_mmwr_epiweek(last_data_date_overall + 1) +
+          lubridate::ddays(
+            (floor(n_forecast_days / nhsn_step_size) - 1) * nhsn_step_size
```
dylanhmorris, line 387: just use integer division?

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -856,7 +860,9 @@ def batch_colvolve_fn(m):
         site_log_ww_conc = DistributionalVariable(
```
dylanhmorris, line 860: Not for this PR, but I think we can now remove this (as it's more consistent with the other signals just to handle this via `Predictive`

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -386,7 +386,9 @@ def sample(
         if (
             model_t_observed is None
         ):  # True for forecasting/posterior prediction
-            which_obs_ed_visits = jnp.arange(potential_latent_ed_visits.size)
+            which_obs_ed_visits = jnp.arange(
+                -model_t_first_latent_ed_visit, potential_latent_ed_visits.size
```
dylanhmorris, line 390: This shouldn't have been reverted, should it?

dylanhmorris (commented):
A few questions @sbidari

hewr/R/process_state_forecast.R
```diff
@@ -372,6 +373,25 @@ process_pyrenew_model <- function(model_run_dir,
...
+  mismatch <- model_samples_tidy |>
+    dplyr::group_by(.variable) |>
+    dplyr::summarise(predicted_last_date = max(date)) |>
+    dplyr::mutate(
+      expected_last_date = dplyr::case_when(
+        stringr::str_ends(.variable, "ed_visits") ~
+          last_data_date_overall + n_forecast_days,
+        .variable == "site_level_log_ww_conc" ~ last_data_date_overall,
+        stringr::str_ends(.variable, "hospital_admissions") ~
+          forecasttools::ceiling_mmwr_epiweek(last_data_date_overall + 1) +
```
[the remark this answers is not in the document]
sbidari: Ensures first forecast date is the following Saturday if `last_data_date` is a Saturday

pyrenew_hew/pyrenew_hew_model.py
sbidari: I believe this is correct indexing since first entry of `potential_latent_ed_visits` corresponds to `model_t_first_latent_ed_visit` but we want posterior prediction to start from model time `t0`.

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -856,7 +860,9 @@ def batch_colvolve_fn(m):
         site_log_ww_conc = DistributionalVariable(
             "site_log_ww_conc",
             dist.Normal(
-                loc=expected_obs_viral_genomes[:, lab_site_to_subpop_map]
+                loc=expected_obs_viral_genomes[
+                    -model_t_first_latent_viral_genome:, lab_site_to_subpop_map
```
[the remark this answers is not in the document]
sbidari: Similar reasoning as above!

hewr/R/process_state_forecast.R
sbidari: https://github.com/CDCgov/pyrenew-hew/pull/473/commits/c1447f7ad193aa8bd177699dbadd185edf642492

dylanhmorris: Would this not achieve the same thing more simplfy?

I think it should be floor because I believe it throws out incomplete epiweeks but if it doesn't it should still be ceiling
```suggestion
        stringr::str_ends(.variable, "hospital_admissions") ~
          forecasttools::floor_mmwr_epiweek(last_data_date_overall + lubridate::days(n_forecast_days))
```

hewr/R/process_state_forecast.R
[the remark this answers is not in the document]
dylanhmorris: got it. now that I understand what you're after, I have a countersuggestion below https://github.com/CDCgov/pyrenew-hew/pull/473/files/1c2083212d25c330bbe7548878ccf3346bd98805#r2074133993

pyrenew_hew/pyrenew_hew_model.py
dylanhmorris: I don't think this will work as intended due to the minus sign.
```
            which_obs_ed_visits = jnp.arange(
                -model_t_first_latent_ed_visit, potential_latent_ed_visits.size
 ```
Say `model_t_first_latent_ed_visit = 5` and `potential_latent_ed_visits.size = 10` and This will be a sequence of indices [-5, -4, ... 9], which is not what you want. I think you want
[5, 6, ... 9] 

That said, I think we should generate the full set of posterior predictions and then letter index to model time later if needed, rather than truncating here.

**Review by dylanhmorris — COMMENTED** [review]
A few more comments @sbidari

File: pyrenew_hew/pyrenew_hew_model.py (hunk 3 above)
Comment (sbidari) [reply pyrenew_hew/pyrenew_hew_model.py:390]: I should have also mentioned `model_t_first_latent_ed_visit` is negative as it is the sum of `ed_visit_offset` and `n_initialization_points` (this is negative and greater in magnitude than `offset`).
```
        model_t_first_latent_ed_visit = (
            ed_visit_offset + model_t_first_latent_infection
        )
```
So in the example above, we'd have `model_t_first_latent_ed_visit = -5` and get the sequence of indices `[5, 6, ... 9]`

Comment (sbidari) [reply pyrenew_hew/pyrenew_hew_model.py:390]: I think we are generating the full posterior prediction but truncating to exclude initialization period, no?

I see the benefits of postponing the truncation until post-processing (refit the model vs just re-do post processing). But this will require the offset values for different signals from the initialization to be available during post postprocessing.

File: hewr/R/process_state_forecast.R (hunk 1 above)
Comment (sbidari) [reply hewr/R/process_state_forecast.R]: The issue with this implementation is that we need `forecasttools::floor_mmwr_epiweek` if `last_data_date_overall` is Friday and `forecasttools::ceiling_mmwr_epiweek` if Saturday.

An example with `n_forecast_days= 28`:

```
last_data_date_overall = as.Date("2025-01-04") #Saturday
forecasttools::ceiling_mmwr_epiweek(last_data_date_overall + lubridate::days(n_forecast_days))
[1] "2025-02-01"

last_data_date_overall = as.Date("2025-01-03") #Friday
forecasttools::ceiling_mmwr_epiweek(last_data_date_overall + lubridate::days(n_forecast_days))
[1] "2025-02-01"   #29 days

pyrenew_hew/pyrenew_hew_model.py
dylanhmorris: Got it. Thanks for the explanation @sbidari! Recording that the goal here is have the posterior predictions start from model t0, (instead of `model_t_first_latent_{signal}`, which is _before_ model t0). Longer term we may do more general posterior prediction by default and then subsample later, but that will wait for another PR (in which we carry model time / human time mapping forward into postprocessing)

pyrenew_hew/pyrenew_hew_model.py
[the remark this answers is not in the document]
dylanhmorris: Resolved as above.

hewr/R/process_state_forecast.R
sbidari: resolved in https://github.com/CDCgov/pyrenew-hew/pull/473/commits/372190bea64efa8b41e91122e994658b2612dacf

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -856,7 +866,9 @@ def batch_colvolve_fn(m):
...
```
dylanhmorris, line 870: explanatory comment here as well?

pyrenew_hew/pyrenew_hew_model.py
```diff
@@ -536,20 +543,30 @@ def calculate_weekly_hosp_indices(
...
+                    "Not all observed or predicted hospital admissions are on Saturdays."
+                )
+            which_obs_weekly_hosp_admissions = (
+                model_t_observed - model_t_first_pred_admissions
+            ) // 7
+        else:
+            which_obs_weekly_hosp_admissions = jnp.arange(n_datapoints)
+            if model_t_first_pred_admissions < 0:
+                which_obs_weekly_hosp_admissions = (
+                    which_obs_weekly_hosp_admissions[
```
dylanhmorris, line 566: Explanatory comment that this make it the first epiweek ends after t0 for posterior prediction?

PR #484, opened 2025-05-09
Replace values of wastewater concentration lower than LOD with 0.5*LOD

### LINKED ISSUE #423: Prevent log_genome_copies_per_ml from being -Infinity
This will error if/when `log_genome_copies_per_ml` contains `-Infinity`

_Originally posted by @sbidari in https://github.com/CDCgov/pyrenew-hew/pull/388#discussion_r2042345557_

Then we can remove all instances of `str_replace_all("-Infinity", "null")`

sbidari:
closes #423

### DIFF
--- a/pipelines/prep_ww_data.py
+++ b/pipelines/prep_ww_data.py
@@ -64,7 +64,8 @@ def clean_nwss_data(nwss_data):
                     ]
                 )
             )
-            | (pl.col("quality_flag").is_null())
+            | (pl.col("quality_flag").is_null()),
+            pl.col("pcr_target_avg_conc") >= 0,
         )
     ).drop(["quality_flag", "pcr_target_units"])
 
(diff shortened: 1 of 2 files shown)


pipelines/prep_ww_data.py
```diff
@@ -80,6 +80,13 @@ def clean_nwss_data(nwss_data):
         ]
     )
 
+    # If pcr_target_avg_conc is zero, replace with 0.5*LOD
+    nwss_subset_clean = nwss_subset_clean.with_columns(
+        pcr_target_avg_conc=pl.when(pl.col("pcr_target_avg_conc") <= 0)
```
dylanhmorris, line 85: I think I'd first filter out negative values (since they could be errors) and then set anything non-negative but sub-LOD to 0.5 * LOD.

dylanhmorris: IOW we probably don't want to classify -99 as "<LOD"

dylanhmorris (changes requested):
One minor point.

pipelines/prep_ww_data.py
sbidari: I am sorry, I forgot we'd decided to set all sub-LOD to 0.5*LOD and not just the zero values, will change this.

sbidari: changed in https://github.com/CDCgov/pyrenew-hew/pull/484/commits/ece1639975a4eb5ceddc4a385f187db1f0bf1740

dylanhmorris (approved):
LGTM, thanks!

PR #489, opened 2025-05-15
Use `forecasttools::sym_limits()` in `summarize_visualize_scores.R`

dylanhmorris:


dylanhmorris:
Good point, we can probably just remove this now that scoring happens in the minihub etc

damonbayer:
@dylanhmorris I would accept a PR to remove this file, if the mini hub capabilities are sufficient.

damonbayer:
@dylanhmorris It's a bit of a stretch to say backtesting is more supported by this repo than the minihub repo, in my opinion. 
We do not use any of the scoring functionality in this repo on a regular basis (note that [`score_forecast`](https://github.com/CDCgov/pyrenew-hew/blob/0a96e8f3432a7e068736128230c201e05d67b545/pipelines/forecast_state.py#L166) is defined, but never used).
We should either move toward documenting this aspect (scoring / backtesting) of the repo or toward removing these unused capabilities (and migrating to the minihub).

dylanhmorris:
> We should either move toward documenting this aspect (scoring / backtesting) of the repo or toward removing these unused capabilities (and migrating to the minihub).

Good point. I think we should move toward a full backtesting workflow but I'm not sure the minihub is where it will live. I liked your proposal from a while back for general tooling for on-demand hubverse-format forecasts for backtesting. Perhaps in `forecasttools`?

PR #491, opened 2025-05-16
Tweak setup_pool.py

dylanhmorris:
Simpler and clearer to get the compute node identity reference directly from the `CredentialHandler`. I am planning a breaking change to `cfa-azuretools` that will refactor and the rename the `get_compute_node_id_reference()` function to `get_compute_node_identity_reference()`; this change will prevent that one from breaking the pipeline.


sbidari (approved):
LGTM!

PR #492, opened 2025-05-18
Archive `stderr` and `stdout` from batch tasks to blob

### LINKED ISSUE #470: Persist Azure batch job runtime errors
Currently we can't view `stderr.txt` / `stdout.txt` file once the node is inactive. Persisting these files or any other form of error logging would be helpful in debugging

dylanhmorris:
Closes #470

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
     "pyarrow>=18.0.0",
     "pygit2>=1.17.0",
     "tomli-w>=1.1.0",
-    "azuretools",
+    "azuretools @ git+https://github.com/cdcgov/cfa-azuretools",
 ]
 
 [build-system]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -47,3 +47,6 @@ line-length = 79
 
 [tool.ruff.lint]
 select = ["I"]
+
+[tool.uv.sources]
+azuretools = { git = "https://github.com/cdcgov/cfa-azuretools" }
```
damonbayer, line 52: I would prefer to use standard pyproject features over uv-specific features when possible. Is there a benefit to this change?

dylanhmorris: None that I'm aware of. I changed the endpoint a few times (pointed at the dev branch and then updated to point at main) using the `uv add`, which adds git remotes to this entry per https://docs.astral.sh/uv/concepts/projects/dependencies/#adding-dependencies

Will revert.

PR #497, opened 2025-05-20
update azure batch VM size used for pyrenew-hew model runs

sbidari:


dylanhmorris:
I think we can keep the default pool as-is. If we expect to be creating more pools of various sizes, perhaps make the vm a configurable command line arg?

damonbayer:
I think the new pool is already created. We could add this change but comment it out and note that it is for W models.

sbidari:
The new pool is already created, and I don't think we need to create pools often. I like adding a comment that models with W signals need larger vm size.

dylanhmorris:
Can you do that and re-request review @sbidari?

damonbayer (approved):

PR #506, opened 2025-05-22
rename timeseries forecasts

### LINKED ISSUE #185: Rename baseline forecasts
We should give them a more descriptive name other than baseline. Assuming we will have baseline-like forecasts for both ED and Hosp data streams, the model names should also indicate the data stream they use.

sbidari:
closes #185


damonbayer:
Tentative LGTM. Thanks @sbidari! Can you also take care of related PR's in https://github.com/cdcent/cfa-stf-minihub/ (renaming the existing models) and https://github.com/cdcent/cfa-hubverse-processor (translating from the new model names to the new model names in the hub ([here](https://github.com/cdcent/cfa-hubverse-processor/blob/main/R/write_table_to_local_hub.R)). Please let me know if you need additional guidance.

PR #519, opened 2025-05-30
fix names and paths of timeseries models to ensure inclusion in the hubverse table

sbidari:
changes missed in #506

### DIFF
--- a/pipelines/prep_data.py
+++ b/pipelines/prep_data.py
@@ -251,7 +251,7 @@ def process_loc_level_data(
             .replace(_inverse_disease_map),
         )
         .sort(["date", "disease"])
-        .collect(streaming=True)
+        .collect(engine="streaming")
     )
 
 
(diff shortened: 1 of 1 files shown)


hewr/R/to_hub_quantile_table.R
```diff
@@ -55,7 +55,7 @@ to_hub_quantile_table <- function(model_batch_dir) {
     quantiles_paths <- fs::dir_ls(
       model_run_dir,
       recurse = TRUE,
-      glob = "*_quantiles.parquet"
+      glob = "*_quantiles_e.parquet"
```
dylanhmorris, line 58: I think we want this in case we add other signal quantile forecasts in the future 
```suggestion
      glob = "*_quantiles_*.parquet"
```

dylanhmorris (approved):
Thanks @sbidari! One question/suggestion.

Also, looks like we need to update a line of code elsewhere to be compatible with the latest `polars`. I think that's a small enough change that it could be in this PR or it could be a separate one. Could you do?

sbidari:
> Also, looks like we need to update a line of code elsewhere to be compatible with the latest `polars`. I think that's a small enough change that it could be in this PR or it could be a separate one. Could you do?

Looking into it!

PR #520, opened 2025-05-30
remove deprecated polars arg

sbidari:


pipelines/prep_data.py
```diff
@@ -307,7 +307,7 @@ def aggregate_facility_level_nssp_to_loc(
         .rename({"reference_date": "date"})
         .sort(["date", "disease"])
         .select(["date", "geo_value", "disease", "ed_visits"])
-        .collect(streaming=True)
+        .collect(engine="streaming")
         # setting streaming = True explicitly
```
dylanhmorris, line 311: Change this comment to reflect the new arg?

dylanhmorris (approved):
Thanks!

pipelines/prep_data.py
sbidari: done in https://github.com/CDCgov/pyrenew-hew/pull/520/commits/2a19942497135a639c896dfa1f7b4fb18de96081

PR #524, opened 2025-06-03
separate timeseries model from pyrenew

sbidari:


.github/workflows/pipeline-run-check.yaml
```diff
@@ -72,9 +108,16 @@ jobs:
           name: test-data
           path: ${{ env.DATA_DIR }}
 
+      - name: Download timeseries output
```
dylanhmorris, line 111: do this step only for the e models with an if?

pipelines/batch/setup_job.py
```diff
@@ -162,30 +172,57 @@ def main(
...
+            f"--model-letters {model_letters} "
+            "--eval-data-path "
+            "nssp-etl/latest_comprehensive.parquet"
+            "'"
+        )
+    else:
+        raise ValueError(
+            f"Unsupported model family: {model_family}. "
+            "Supported values are 'pyrenew' and 'timeseries'."
+        )
```
dylanhmorris, line 225: Looks to me as though the call is identical apart from the script that is called. define a `run_script` variable in the if/else and then do 

```
base_call = (
   f"uv run python pipelines/{run_script}"
```
 etc?

pipelines/forecast_loc.py
```diff
@@ -367,9 +349,19 @@ def get_available_nwss_reports(
...
-
     os.makedirs(model_run_dir, exist_ok=True)
 
+    if forecast_ed_visits and not os.path.exists(
+        Path(model_run_dir, "timeseries_e")
+    ):
+        raise ValueError(
+            "timeseries_e model run not found. "
+            "Please ensure that the timeseries forecasts "
+            "are generated before fitting Pyrenew models. "
```
dylanhmorris, line 360:
```suggestion
            "are generated before fitting Pyrenew-*E* models. "
```

pipelines/forecast_loc.py
```diff
@@ -367,9 +349,19 @@ def get_available_nwss_reports(
...
     model_run_dir = Path(model_batch_dir, "model_runs", loc)
-
     os.makedirs(model_run_dir, exist_ok=True)
 
+    if forecast_ed_visits and not os.path.exists(
+        Path(model_run_dir, "timeseries_e")
+    ):
+        raise ValueError(
+            "timeseries_e model run not found. "
+            "Please ensure that the timeseries forecasts "
```
dylanhmorris, line 359:
```suggestion
            "Please ensure that the timeseries forecasts for the E signa,l"
```

pipelines/forecast_loc.py
dylanhmorris, line 1: rename to `forecast_pyrenew.py` to parallel `forecast_timeseries.py`?

sbidari: done in https://github.com/CDCgov/pyrenew-hew/pull/524/commits/49d3d839dc7eca3f0605618b3a87b2074762dfc3

.github/workflows/pipeline-run-check.yaml
dylanhmorris: The non-E models need those to fits to proceed (e.g. you should be able to do HW without a timeseries_e). Not downloading the timeseries fits in this test will help us check that the non-E models are robust to timeseries_e being absent.

pipelines/forecast_loc.py
damonbayer: @sbidari This file is still in the PR

damonbayer (approved):
Thanks @sbidari! I think this can be merged after open comments are addressed.

pipelines/generate_epiweekly_data.R
```diff
@@ -82,24 +82,27 @@ convert_daily_to_epiweekly <- function(
...
 # Create a parser
 p <- arg_parser("Create epiweekly data") |>
   add_argument(
     "model_run_dir",
     help = "Directory containing the model data and output."
+  ) |>
+  add_argument(
+    "--data-names",
+    help = "Whitespace-separated list of data file names to process.",
+    default = "combined_training_data.tsv combined_eval_data.tsv"
```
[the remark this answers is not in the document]
sbidari: I thought I could get away without having to generate eval data for timeseries fit so I made this change to only convert training data into epiweekly. Didn't end up going that route but thought I can leave this change as it doesn't break anything. (Also mentioned it in the PR description https://github.com/CDCgov/pyrenew-hew/pull/524#issue-3115744512)

pipelines/forecast_loc.py
sbidari: My bad! Removed in https://github.com/CDCgov/pyrenew-hew/pull/524/commits/c7c771eb7d8efaed446a31ef8b870d8781acb1d7

pipelines/generate_epiweekly_data.R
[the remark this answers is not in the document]
sbidari: https://github.com/CDCgov/pyrenew-hew/pull/524/commits/2079bc3953d3bbad3b09aa056d6ee8449723fd78

Issue #477, opened 2025-05-01
Make test data more realistic


damonbayer:
Our unrealistic test data makes for poor model fits, which makes diagnosing pipeline problems unnecessarily difficult. This could be resolved by simulating test data from a PyRenew model.

Issue #313, opened 2025-01-28
Enable test-forecast-state to work without requiring internet connection


damonbayer:
We should create fake NHSN data to use in testing, so that we do not have to pull it from the internet.

sbidari:
Could you provide more clarification here. 

NHSN data is publicly available, so we likely don’t need to generate fake data. 
We could save a local copy of the data for offline access, but that approach would differ from the standard model run pipeline and not be a _true_ end-to-end test

damonbayer:
I think I made this issue when NHSN data was unavailable briefly (maybe for a few hours), which made it impossible for us to run the testing pipeline. I am interested in saving a local copy for offline access, just to make testing offline possible/faster, but it's not a high priority. I agree it is not a "true" end-to-end test.

PR #530, opened 2025-06-06
Simulate test data from PyRenew Models

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/313.
Closes https://github.com/CDCgov/pyrenew-hew/issues/477.

- [x] Adds `nhsn_data_path` argument to various places to allow loading of a local file, rather than pulling NHSN data from the API.
- [x] Data is now simulated from the prior predictive of a PyRenew-HEW model in `pipelines/generate_test_data.py`
- [x] Removes code and data previously used to simulate tests data.


damonbayer:
@sbidari The new time series jobs appear to be very fragile. Do you know what's going on?

sbidari:
> @sbidari The new time series jobs appear to be very fragile. Do you know what's going on?

Can you elaborate?

damonbayer:
> > @sbidari The new time series jobs appear to be very fragile. Do you know what's going on?
> 
> Can you elaborate?

Actually, maybe they're totally broken. The PyRenew models complain that they can't find the time series results, even though the time series models ran successfully.

dylanhmorris (approved):
Thanks @damonbayer!

PR #553, opened 2025-06-20
swap to arrow

damonbayer:


dylanhmorris:
Should we just use `forecasttools::read_tabular_file` here so that we can swap back in a more DRY way later?

PR #577, opened 2025-07-10
update read_and_combine_data

### LINKED ISSUE #503: Multiple values of wastewater concentration for a given `lab_site_index` and `date` causes model with W signal to fail during postprocessing

sbidari:
closes #503


damonbayer:
Can we modify the test pipeline to demonstrate that this is fixed?

PR #599, opened 2025-07-23
Update azure_command_center.py

sbidari:


damonbayer:
I'm going to move the command center inside the `batch` folder for compatibility with https://github.com/CDCgov/pyrenew-hew/pull/604

PR #587, opened 2025-07-16
Jkislin makefile setup job enhancements

jkislin:

PR #601, opened 2025-07-24
Submitting jobs to Azure Batch via Cron Jobs

jkislin:
Very much a draft. Lots of stress testing to do. Workflow does not yet work as written.

Steps:
1. Make use of an OIDC Federated Identity in the GIthub Actions workflow, on the STFT Service Principal. 
2. Use the cfa-runner-action.
3. Run the uv commands to launch setup_job.py. 
4. If postprocessing is needed, we'll need to refactor to use the Azure SDK for storage instead of blobfuse, most likely.
   - If this is truly a dealbreaker, we can look into scheduling this pipeline fully _within_ Azure. 

This PR partially assumes that #587 will be eventually merged, at least the `setup_job.py` enhancements therein.


jkislin:
Holding on this - function apps might be the way to go if we just want direct triggers.

jkislin:
**Complete:**
1. Implemnt oidc token.
1. Implement cdcgov/cfa-runner-action.
1. Create scaffolding for all jobs. 
2. Layout general job dependency graph using `needs`.
3. Test dry runs of each job. 

**TODO:**
1. Fix ${{ matrix.model }} interpretation in hw/hew matrices.
1. Complete the "check data availability" step using logic from the Azure Command Center.
2. Implement this logic to spawn jobs conditionally. 
3. Test runs without dry_run flag, including azure environment variables. 
4. Confim cron syntax.

PR #604, opened 2025-07-24
Restrict Actions Workflow Triggers to Only Necessary Paths

jkislin:
Doesn't necessarily make sense that `README.md` edits, etc., should launch intensive test workflows like `pipeline-run-check.yaml`, etc. I'm finding these trigger-happy workflows to be cumbersome/cluttering when trying to test individual Github Actions changes.

We do this in many places elsewhere in CFA - restricting workflow runs to only trigger when relevant paths update. 

With this PR explicitly defined which paths should launch the ~three~ two most-trigger-happy workflows:
- `pipeline-run-check.yaml`
- `r-cmd-check.yaml`
- ~`containers.yaml`~

I'll mark this ready for review after taking a few more passes. (I may have been too aggressive in some places and not aggressive enough in others.)

I've also used `paths` instead of `paths-ignore` - we may wish to use the latter. Open to feedback! Let me know if this isn't helpful.


damonbayer (changes requested):
Concerned about the container workflow, but the rest look good.

.github/workflows/containers.yaml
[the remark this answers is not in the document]
jkislin: I'll revert. The main culprit is definitely `pipeline-run-check.yaml` anyway.

.github/workflows/pipeline-run-check.yaml
```diff
@@ -2,8 +2,33 @@ name: Pipeline Run Check
 
 on:
   pull_request:
+    branches: [main]
+    paths:
+      # Explicitly include paths that should trigger the workflow
+      - 'pipelines/**'
+      - 'pyrenew_hew/**'
+      - '.github/workflows/pipeline-run-check.yaml'
+      - 'Containerfile'
```
damonbayer, line 11: containerfile shouldn't impact the pipeline at all.

damonbayer (approved):
One small suggestion.

.github/workflows/pipeline-run-check.yaml
```diff
@@ -2,8 +2,31 @@ name: Pipeline Run Check
...
+      # Explicitly include paths that should trigger the workflow
+      - 'pipelines/**'
+      - 'pyrenew_hew/**'
+      - '.github/workflows/pipeline-run-check.yaml'
+      - '.github/actions/**'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - '.python-version'
+      # Explicitly exclude batch script changes from triggering this workflow
+      - '!pipelines/batch/**'
```
damonbayer, line 16:
```suggestion
      - '!pipelines/batch/**'
      - '!pipelines/azure_command_center.py`
```

.github/workflows/pipeline-run-check.yaml
```diff
@@ -2,8 +2,31 @@ name: Pipeline Run Check
...
```
damonbayer, line 29:
```suggestion
      - '!pipelines/batch/**'
      - '!pipelines/azure_command_center.py`
```

Issue #573, opened 2025-07-07
Atomic data preparation


dylanhmorris:
> A lot of the repetition is in data preparation steps and can be omitted by separating data preparation from model runs. We can create `prep_data` or similarly named function that prepares data for all possible model runs at once instead of re-writing data directory during each model run.

 _Originally posted by @sbidari in [#549](https://github.com/CDCgov/pyrenew-hew/issues/549#issuecomment-3045761455)_

One question/thought: we want a single prepped dataset per location and as-of date, across all signals, yes? 

Given that:
  - Ideally we shouldn't need, e.g. the latest H data to be available in order to prep data for E models
  - But we _do_ want to overwrite that dataset safely once that H data _is_ available.

I think a simple overwrite can work in our use-case provided we're careful. Main thing I want to avoid: 
- Prep dataset `data_for_model_fit.json` and fit E model before H data released
- Rerun prep data once H data released, dataset `data_for_model_fit.json` is overwritten
- Existing E model fit is no longer compatible with the new `data_for_model_fit.json` file (existing can't be postprocessed with new `data_for_model_fit.json`, E model would give different results if you re-fit to the `data_for_model_fit.json`, etc)

dylanhmorris:
Related: #535, #568

damonbayer:
While we're doing this, I think we could strip out all of the logic for getting any dataset that isn't the latest data. It should simplify the code quite a bit.

PR #606, opened 2025-07-24
prune data_for_model_fit.json

sbidari:
- [x] prune `data_for_model_fit.json` with only the info necessary to create `pyrenewHEWData`
- [x] add step sizes for all training data to `pyrenewHEWData`
- [x] Change `dates_observed_{signals}` to be np.array over pl.Series (Date inputs to `pyrenewHEWData` now must be `np.datetime64` if manually passed)
- [x] Updated tests
- [x] Incorporate `filter_params` functionality in `get_pmfs`
- [x] add a method to construct pyrenewHEWData from json file
- [x]  #607
- [x] update test_build_pyrenew_model 
- [x] Added a new class `PyrenewHEWParam` for model parameters passed to `build_pyrenew_hew_model` and a `from_json` method for construction. Separates `PyrenewHEWData` for model run from `PyrenewHEWParam` for model build.
- [x] added build_pyrenew_hew_model_from_dir to pipelines.utils


pipelines/forecast_pyrenew.py
```diff
@@ -179,6 +181,57 @@ def create_hubverse_table(model_fit_path):
     return None
 
 
+def get_pmfs(param_estimates: pl.LazyFrame, loc_abb: str, disease: str):
```
damonbayer, line 184: See also `filter_params` here: https://github.com/cdcent/stf-pyrenew-added-value/blob/main/src/stf_pyrenew_added_value/pull_data_as_of.py#L75

I would rather import this functionality from `pyrenew-hew`, so please take a look at that code and make sure this update is the best of both worlds.

pipelines/forecast_pyrenew.py
```diff
@@ -175,6 +177,141 @@ def create_hubverse_table(model_fit_path):
...
+
+
+def get_pmfs(
+    param_estimates: pl.LazyFrame,
+    loc_abb: str,
+    disease: str,
+    as_of: dt.date = None,
+    reference_date: dt.date = None,
+    allow_missing_right_truncation: bool = False,
+) -> tuple[list, list, list]:
```
damonbayer, line 207: Return a dict instead, so these can be named.

pyrenew_hew/utils.py
```diff
@@ -310,27 +310,28 @@ def flags_from_pyrenew_model_name(model_name: str) -> dict[str, bool]:
...
 def build_pyrenew_hew_model(
-    model_data: dict,
     priors: dict,
+    pop_fraction: ArrayLike,
+    population_size: float,
+    generation_interval_pmf: ArrayLike,
+    inf_to_hosp_admit_lognormal_loc: ArrayLike,
+    inf_to_hosp_admit_lognormal_scale: ArrayLike,
+    inf_to_hosp_admit_pmf: ArrayLike,
+    right_truncation_pmf: ArrayLike = None,
```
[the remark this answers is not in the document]
damonbayer: I *think* I would prefer the previous approach where we pack all of these parameters into `PyRenewHEWData`. Conceptually, maybe some of its elemets are better described as "parameters". I would be open to having a bit more structure, but I think it's important to be able to go from "model config" to "runnable model" with very few function arguments.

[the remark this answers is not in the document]
sbidari: The PMFs were not included in `PyrenewHEWData` but were stored in the `data_for_model_fit.json`. I removed the pmfs from the json file to only include information in `PyrenewHEWData` in `data_for_model_fit.json` (based on the issue description). I think there are few different ways to handle this:
1. revert the change to `PyrenewHEWData` and `data_for_model_fit.json` to back to how it was
2. Add the pmfs and related objects to `PyrenewHEWData` so that `data_for_model_fit.json` only the information needed to craete `PyrenewHEWData` 
3. Store the pmfs in some config type file, maybe we can add these to the priors (also passed to `build_pyrenew_hew_model`)?

Storing the pmfs as `PyrenewHEWData` feels incorrect, I think option 1 (reverting to how it was) or 3 would be better. Lmk which one you prefer

pyrenew_hew/pyrenew_hew_param.py
```diff
@@ -0,0 +1,58 @@
...
+    Class for holding PyrenewHEW model parameters
+    """
+
+    population_size: int
+    pop_fraction: ArrayLike
+    generation_interval_pmf: ArrayLike
+    inf_to_hosp_admit_lognormal_loc: float
+    inf_to_hosp_admit_lognormal_scale: float
+    inf_to_hosp_admit_pmf: ArrayLike
+    right_truncation_pmf: ArrayLike = None
```
damonbayer, line 22: Why is this allowed to be `None`?

damonbayer: Should we default to a `[1]` array instead?

sbidari: I forgot I set default value to be `[1]` and was trying to accommodate missing right truncation pmfs. Will remove!

sbidari: https://github.com/CDCgov/pyrenew-hew/pull/606/commits/768e60aa6996fc2e0ac4080f2210969cf034e0c0

pipelines/forecast_pyrenew.py
sbidari: https://github.com/CDCgov/pyrenew-hew/pull/606/commits/4329e89be0c234fbbb734ab2b988401c4e5e02a8

pyrenew_hew/utils.py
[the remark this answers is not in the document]
sbidari: added `PyrenewHEWParam` for arguments to `build_pyrenew_hew_model` so these can be separated from `PyrenewHEWData` passed during model run https://github.com/CDCgov/pyrenew-hew/pull/606/commits/3eab35b3b6cd204306d6dfad0f5b71493f53f18c

tests/test_build_pyrenew_model.py
damonbayer, line 1: Restore tests for `build_pyrenew_hew_model`? Here or in `tests/test_pyrenew_hew_param.py`?

sbidari: I moved the tests to `tests/test_pyrenew_hew_data.py`. The test assertions were all on PyrenewHEWData so that seemed like a better place.

pipelines/fit_pyrenew_model.py
```diff
@@ -30,10 +33,21 @@ def fit_and_save_model(
...
+        / "data"
+        / "data_for_model_fit.json",
+        fit_ed_visits=fit_ed_visits,
+        fit_hospital_admissions=fit_hospital_admissions,
+        fit_wastewater=fit_wastewater,
+    )
+    model_params = PyrenewHEWParam.from_json(
+        Path(model_run_dir) / "model_params.json"
+    )
+    my_model = build_pyrenew_hew_model(
```
damonbayer, line 48: I think it would be useful to remove the `get_priors_from_dir` function and bundle all of this in a `build_pyrenew_hew_model_from_dir` function, which has the relative paths as arguments and good defaults.

pipelines/forecast_pyrenew.py
```diff
@@ -403,12 +407,19 @@ def get_available_nwss_reports(
         report_date=report_date,
         first_training_date=first_training_date,
         last_training_date=last_training_date,
-        param_estimates=param_estimates,
         model_run_dir=model_run_dir,
         logger=logger,
         credentials_dict=credentials_dict,
         nhsn_data_path=nhsn_data_path,
     )
+    process_and_save_loc_param(
```
damonbayer, line 415: Restore `process_and_save_loc`, but refactor it to call `process_and_save_loc_data` and `process_and_save_loc_param`?

pyrenew_hew/pyrenew_hew_data.py
```diff
@@ -17,21 +20,26 @@ def __init__(
...
-        first_hospital_admissions_date: datetime.date = None,
-        first_wastewater_date: datetime.date = None,
+        first_ed_visits_date: np.datetime64 = None,
+        first_hospital_admissions_date: np.datetime64 = None,
+        first_wastewater_date: np.datetime64 = None,
         right_truncation_offset: int = None,
-        pop_fraction: ArrayLike = None,
         n_ww_lab_sites: int = None,
         lab_site_to_subpop_map: ArrayLike = None,
         population_size: int = None,
```
damonbayer, line 35: add `| None` to the type hint for all of these.

damonbayer (changes requested):
Really nice overall, but there are a few places I think we can simplify.

pipelines/forecast_pyrenew.py
sbidari: These will be moved to `prep_data.py` to reduce redundancies between `forecast_pyrenew.py` and `forecast_timeseries.py`. I would rather make changes to these in #573 than now

pipelines/prep_data.py
```diff
@@ -322,64 +323,149 @@ def get_loc_pop_df():
...
+    ).collect()
+
+    generation_interval_pmf = _validate_and_extract(
+        generation_interval_df, "generation_interval"
     )
 
+    delay_df = filtered_estimates.filter(
+        pl.col("parameter") == "delay"
+    ).collect()
+    delay_pmf = _validate_and_extract(delay_df, "delay")
```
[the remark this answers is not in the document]
sbidari: https://github.com/CDCgov/pyrenew-hew/pull/606/commits/51d0fe14154cde95fc779bf94faab7cd77217e31

Issue #591, opened 2025-07-18
Remove functionality for pulling non-latest data.


damonbayer:
> While we're doing this, I think we could strip out all of the logic for getting any dataset that isn't the latest data. It should simplify the code quite a bit. 

 _Originally posted by @damonbayer in [#573](https://github.com/CDCgov/pyrenew-hew/issues/573#issuecomment-3050289724)_

PR #611, opened 2025-07-30
atomic data preparation

sbidari:
Major changes:

- [x] Move all data preparation logic from `forecast_pyrenew.py` and `forecast_timeseries.py` to `prep_data.py`
- [x] Add data preparation as a separate job type in the Azure command center and CI runs
- [x] Remove generation of eval data 
- [x]  #591


tests/test_pyrenew_hew_data.py
```diff
@@ -130,6 +130,12 @@ def test_to_forecast_data(
         ).weekday()
         == 5
     )
+    assert (
+        forecast_data.first_hospital_admissions_date.astype(
+            dt.datetime
+        ).weekday()
+        == 5
+    )
```
sbidari, line 138:
```suggestion
```

pipelines/forecast_pyrenew.py
```diff
@@ -234,149 +170,18 @@ def main(
...
-            )
-    else:
-        loc_level_nwss_data = None
 
-    param_estimates = pl.scan_parquet(Path(param_data_dir, "prod.parquet"))
-
-    model_batch_dir_name = (
-        f"{disease.lower()}_r_{report_date}_f_"
-        f"{first_training_date}_t_{last_training_date}"
+    (_, _, model_run_dir) = get_training_dates_and_model_dir(
```
damonbayer, line 176: Seems like we could remove a lot of vafiables throughout if we just passed `model_run_dir` to `main` instead.

sbidari: This is accomplished in https://github.com/CDCgov/pyrenew-hew/pull/611/commits/51283f432f63db5276b4a5826166ac95cae630b7 and https://github.com/CDCgov/pyrenew-hew/pull/611/commits/141da47847533c8565d21c9c7ba272214ad2600b.

pipelines/batch/setup_job.py
```diff
@@ -453,14 +494,14 @@ def style_bool(val):
...
     parser.add_argument(
-        "--model-family",
+        "--run-script",
         type=str,
         help=(
-            "Model family to use for the job. "
-            "Supported values are 'pyrenew' and 'timeseries'. "
-            "Default 'pyrenew'."
+            "The python script to run. "
+            "Supported values are 'forecast_pyrenew', 'forecast_timeseries', and 'prep_data'. "
```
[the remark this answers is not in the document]
sbidari: https://github.com/CDCgov/pyrenew-hew/pull/611/commits/ed06a7ac78b44b065e2d040e17189b071da5264d

pipelines/azure_command_center.py
```diff
@@ -163,6 +174,30 @@ def ask_about_reruns():
...
 
 
+def do_data_prep_reruns(
+    locations_include: list[str] | None = None,
+    e_exclude_last_n_days: int = 1,
+    h_exclude_last_n_days: int = 1,
+    append_id: str = "",
+):
+    if e_exclude_last_n_days == 1:
+        print("Skipping data preparation re-run due to E")
```
dylanhmorris, line 184: This assumes 1 will always be the default, which is not necessarily the case. Especially as we are moving toward a flatter output / data prep structure. I think this can probably be omitted.

dylanhmorris: I see a similar code pattern  in the actual timeseries / pyrenew re-run function, and I'd favor removing it there, too, but maybe I am missing something. @damonbayer can you give insight?

dylanhmorris: In any case, given I can now see that this assumptions occurs throughout the `azure_command_center.py`, I think it should be addressed (if at all) all at once in a separate PR.

damonbayer: > I see a similar if/else str in the actual timeseries / pyrenew re-run function, and I'd favor removing it there, too, but maybe I am missing something. @damonbayer can you give insight?

You could parameterize the default arg. The idea is to not refit models that have already been fit. We always fit the default model before rerunning any models.

pipelines/prep_data.py
```diff
@@ -22,10 +25,28 @@
...
+        ).date()
+        for f in Path(data_dir).glob(f"{glob_pattern}*")
+    ]
+
 
 def get_nhsn(
-    start_date: datetime.date,
-    end_date: datetime.date,
+    start_date: dt.date,
+    end_date: dt.date,
```
dylanhmorris, line 49: Good catch, thanks!

pipelines/prep_data.py
```diff
@@ -368,12 +411,12 @@ def get_pmfs(
...
         Name of the disease.
 
-    as_of : datetime.date, optional
+    as_of : dt.datetime.date, optional
         Date for which parameters must be valid
         (start_date <= as_of <= end_date). Defaults
         to the most recent estimates.
 
-    reference_date : datetime.date, optional
+    reference_date : dt.datetime.date, optional
```
dylanhmorris, line 419:
```suggestion
    reference_date : dt.date, optional
```

pipelines/prep_data.py
```diff
@@ -637,3 +678,284 @@ def process_and_save_loc_param(
...
+                & (pl.col("date") >= first_training_date)
+            )
+            .collect()
+        )
+        loc_level_nwss_data = preprocess_ww_data(nwss_data_cleaned)
+    else:
+        raise ValueError(
+            "NWSS data not available for the requested report date "
+            f"{report_date}"
+        )
```
dylanhmorris, line 810: Given that other parts of the code now work as expected with empty NWSS data frames, should this become a warning rather than an error? That way if our NWSS pipeline ever breaks we can still prep data for the non-W models. Thoughts @sbidari @damonbayer?

pipelines/prep_data.py
```diff
@@ -609,11 +649,12 @@ def process_and_save_loc_param(
         else:
             pop_fraction = subpop_sizes / sum(subpop_sizes)
 
+    param_estimates = pl.scan_parquet(Path(param_data_dir, "prod.parquet"))
     pmfs = get_pmfs(
         param_estimates=param_estimates,
         loc_abb=loc_abb,
         disease=disease,
-        right_truncation_required=fit_ed_visits,
+        right_truncation_required=False,
```
dylanhmorris, line 657: See above on proposed argument change.
```suggestion
        right_truncation_required=right_truncation_required,
```

pipelines/batch/setup_job.py
```diff
@@ -204,46 +213,46 @@ def main(
...
+            f"--exclude-last-n-days {exclude_last_n_days} "
         )
-    elif model_family == "timeseries":
-        run_script = "forecast_timeseries.py"
-        additional_args = ""
     else:
-        raise ValueError(
-            f"Unsupported model family: {model_family}. "
-            "Supported values are 'pyrenew' and 'timeseries'."
+        additional_args = (
```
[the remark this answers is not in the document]
sbidari: I think we don't the raise error here as the CL arg for `setup_job.py` can only take one of the three accepted values. https://github.com/CDCgov/pyrenew-hew/pull/611#discussion_r2257722790

dylanhmorris (approved):
Thanks, @sbidari! Some optional suggestions.

pipelines/prep_data.py
sbidari: As written data preparation looks for nwss data file for a report date, fails if the file is not found but substitutes an empty dataframe if the location is not found in the data file. This seems reasonable to me to ensure models with W signals are indeed fit to W signals (warnings are ignorable) and we can generate NWSS dataset for any date by running the nwss-etl workload. 

However, this setup won't work when we expand to RSV models. I think we should consider adding fit_{signal} flags to prep_data when switching to flat model dirs setup.

sbidari: created #624

PR #629, opened 2025-08-13
Store additional iteration-level NUTS information

### LINKED ISSUE #628: Record the trace of the unnormalized (minus) log posterior density during NUTS runs.
- Should be as simple as setting `extra_fields=('potential_energy',)` at the `MCMC.run()` call (see [here](https://forum.pyro.ai/t/log-probability-of-samples-using-nuts-hmc/2651))

That trace should then get stored in the `sample_stats` group when we convert to arviz idata via `from_numpyro()` (see [official docs example](https://python.arviz.org/en/latest/getting_started/CreatingInferenceData.html#from-numpyro))

dylanhmorris:
Including the number of integration steps and, crucially, the (negative of the) log posterior density

Closes #628

### DIFF
--- a/pipelines/fit_pyrenew_model.py
+++ b/pipelines/fit_pyrenew_model.py
@@ -51,7 +51,7 @@ def fit_and_save_model(
         num_warmup=n_warmup,
         num_samples=n_samples,
         rng_key=rng_key,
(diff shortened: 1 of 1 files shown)


damonbayer:
Have you verified this persists to arviz?

damonbayer (approved):

PR #656, opened 2025-09-03
Prevent command center from forecasting Influenza W models

O957:
This PR:

* [x] Adds a variable for wastewater models to `./pipelines/azure_command_center.py` to prevent the running of model which do not have wastewater data.
* [x] Updates the `DISEASES` to include Influenza, given we are sharing results on Influenza.

### DIFF
--- a/pipelines/azure_command_center.py
+++ b/pipelines/azure_command_center.py
@@ -20,7 +20,7 @@ console = Console()
 
 # TODO: work with specific diseases
 DISEASES = ["COVID-19", "Influenza"]
-WASTEWATER_DISEASES = [d for d in DISEASES if d != "Influenza"]
+WASTEWATER_DISEASES = ["COVID-19"]
 # ND: wastewater data not available
 # TN: wastewater data unusable (dry sludge)
 W_EXCLUDE_DEFAULT = ["US", "TN", "ND"]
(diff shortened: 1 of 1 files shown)


pipelines/azure_command_center.py
```diff
@@ -210,11 +211,14 @@ def do_pyrenew_reruns(
...
-            exclude_last_n_days=h_exclude_last_n_days,
-        )
+
+        if WASTEWATER_DISEASES:
+            fit_pyrenew_hw(
+                append_id=append_id,
+                locations_include=locations_include,
+                exclude_last_n_days=h_exclude_last_n_days,
+                diseases=WASTEWATER_DISEASES,
+            )
```
damonbayer, line 221: I think better to revert these and change the `fit_pyrenew_hw = partial(` definitions.

pipelines/azure_command_center.py
```diff
@@ -225,11 +229,13 @@ def do_pyrenew_reruns(
...
-            locations_include=locations_include,
-            exclude_last_n_days=he_exclude_last_n_days,
-        )
+        if WASTEWATER_DISEASES:
+            fit_pyrenew_hew(
+                append_id=append_id,
+                locations_include=locations_include,
+                exclude_last_n_days=he_exclude_last_n_days,
+                diseases=WASTEWATER_DISEASES,
+            )
```
damonbayer, line 238: revert

damonbayer (approved):

PR #659, opened 2025-09-04
Removing container build cacheing to assess build reliability without it

jkislin:
Running tests. Will request review. It probably makes sense to leave the cache code commented in the script as we may change our mind again :)

### DIFF
--- a/.github/scripts/docker_build_and_push.sh
+++ b/.github/scripts/docker_build_and_push.sh
@@ -8,11 +8,19 @@ BUILDER=docker-container-driver-builder
 docker buildx create --name "$BUILDER" --driver=docker-container || true
 
 # use the registry cache for prior images of the same tag, or the 'latest' tag
+# time docker buildx build --push -t "$IMAGE:$TAG" \
+# 	--builder "$BUILDER" \
+# 	--build-arg "GIT_COMMIT_SHA=$GIT_COMMIT_SHA" \
+# 	--build-arg "GIT_BRANCH_NAME=$GIT_BRANCH_NAME" \
+# 	--cache-from "type=registry,ref=$IMAGE:$TAG-cache" \
+# 	--cache-from "type=registry,ref=$IMAGE:latest-cache" \
+# 	--cache-to "type=registry,ref=$IMAGE:$TAG-cache,mode=max" \
+# 	-f Containerfile .
+
+# attempt without registry cache - we care less about speed 
+# and more about deterministic rebuilds
 time docker buildx build --push -t "$IMAGE:$TAG" \
 	--builder "$BUILDER" \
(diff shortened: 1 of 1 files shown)


dylanhmorris:
> Let me know if there's a way I can help test the functionality of the container, or if that's even necessary given our automated tests. If not necessary, please feel free to review or provide feedback!

I think the best way to test functionality before merging to try running a job in Azure with the [`jk-rm-cache` version of the container](https://github.com/CDCgov/pyrenew-hew/pkgs/container/pyrenew-hew/506496888?tag=jk-rm-cache) via [`setup_job.py](https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/batch/setup_job.py)

jkislin:
> > Let me know if there's a way I can help test the functionality of the container, or if that's even necessary given our automated tests. If not necessary, please feel free to review or provide feedback!
> 
> Thanks, @jkislin! I think the best way to test functionality before merging to try running a job in Azure with the [`jk-rm-cache` version of the container](https://github.com/CDCgov/pyrenew-hew/pkgs/container/pyrenew-hew/506496888?tag=jk-rm-cache) via [`setup_job.py`](https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/batch/setup_job.py). Could you try that and report the results?

I am still getting a Manifest unknown error. 
image

I will investigate more tomorrow.

jkislin:
> > @jkislin I think with these simplifications we can just remove the `docker_build_and_push.sh` script and use the official action, no? [docker/build-push-action](https://github.com/docker/build-push-action?rgh-link-date=2025-09-18T13%3A58%3A57Z)
> > Might resolve the manifest issues and should in general be more robust
> 
> @jkislin What do you think of this?

It's possible that would help, I am noticing that there's either inappropriate (or we just don't understand exactly) redirect behavior when specifying image tags on ghcr.io.

image

Any tag I supply directs here. There's a "latest" meta-tag that applies even if the image pushed wasn't tagged "latest".

I'm now also getting this error on the pool itself. I'm going to see if making "latest" the actual "latest" changes things.

jkislin:
I've refactored for the build-and-push action, and we can toggle cacheing via commenting/uncommenting the final two lines.

.github/scripts/docker_build_and_push.sh
dylanhmorris: I believe this can now be wholly deleted given that the workflow uses `docker/build-push-action@v5` instead of the custom shell script.

Especially given that we have the Git history I prefer deletion to commenting.

dylanhmorris (changes requested):
Thanks, @jkislin! Should be good to go with deletion of old shell script

dylanhmorris (approved):
LGTM!

PR #695, opened 2025-09-23
Issue 690: Deploy `EpiAutoGP`

### LINKED ISSUE #690: `EpiAutoGP` model
Meta-issue for necessary elements for the `EpiAutoGP` model

SamuelBrand1:
This is ready for review now.

I've introduced a new "atomic" model for eventual inclusion into the STF forecast pipeline.

### High level summary

At a high level, `EpiAutoGP` takes in a specific structure of JSON input data along with the kind of command line args you'd expect and returns hubverse (quantile-based) forecasts. This is expecting one JSON file per location/pathogen/target combination so the use-case is aimed at batch processing over the possible combinations of these.

* Running one example is given in the `end-to-end` subdirectory.

* Added `.github/copilot-instructions.md` which contains a summary for agents.

* Detailed usage instructions, input format requirements, and testing strategies for the new EpiAutoGP component in its `README.md`.
* `EpiAutoGP` is excluded from the main container build.

### Outline Input data specifications

- Univariate reports time series with dates, this is not restricted to be integer valued or even non-negative although if negative only use the identity data transformation.
- Nowcasting. nowcast dates overload dates and will be ingested by `NowcastAutoGP`. Arbitrary number of nowcast reports as generative samples of nowcasts, but each nowcast sample must be a vector of the length of the nowcast dates. If both are empty then this lowers to forecasting using `AutoGP` (no nowcasting).

### Not covered in this PR

- **Pre-processing**. This PR is aimed at predictable input->output interface which can be tested as an atomic unit (i.e. not with the differing field names of nssp/nhsn data).
- **Nowcasting model**. This is expected to be available in the input data.
- **Post-processing**. The eventual forecast table will be a vcat over individual forecasts. **Also conversion to FIPS is not done in this PR**.

### More details

**EpiAutoGP Component Introduction:**

* Added the `EpiAutoGP` directory, containing a Julia-based Auto-Gaussian Process modeling package with its own `Project.toml`, `README.md`, and container configuration (`Containerfile`). This includes core modeling code, input/output handling, argument parsing, and a unit test suite.

**Containerization & Integration:**

* Added a dedicated `EpiAutoGP/Containerfile` for building and running the Julia model in isolation from the main Python/R environment, ensuring reproducibility and separation of dependencies.
* Updated the main `Containerfile` to exclude the `EpiAutoGP` directory from the main build context, reinforcing container separation.


SamuelBrand1:
The R formatter is not happy with the example of using `forecasttools` plotting in the end-to-end!

PR #697, opened 2025-09-24
Fix Disease Variable In Pipelines Utilities Script

O957:
This PR:

* [x] Fixes the `disease` argument to `get_all_forecast_dirs` in `pipelines/utils.py` by casting `disease` as a `list`.

The original error:

Screenshot 2025-09-24 at 15 10 24

### DIFF
--- a/pipelines/utils.py
+++ b/pipelines/utils.py
@@ -59,7 +59,7 @@ def parse_model_batch_dir_name(model_batch_dir_name: str) -> dict:
 
 def get_all_forecast_dirs(
     parent_dir: Path | str,
-    diseases: str | set[str],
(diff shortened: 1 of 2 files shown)


pipelines/utils.py
```diff
@@ -59,7 +59,7 @@ def parse_model_batch_dir_name(model_batch_dir_name: str) -> dict:
 
 def get_all_forecast_dirs(
```
damonbayer, line 60: I think it would be better to revert this change and do the list conversion when calling the function.

damonbayer (changes requested):

damonbayer (approved):

Issue #699, opened 2025-09-25
Make testing work offline


damonbayer:
Timeseries test forecast currently requires internet access.

Issue #684, opened 2025-09-19
Each model fit should have its own data directory


damonbayer:
To avoid race conditions and overwriting of the data directory from subsequent model runs, each model fit should have its own data directory, within the current hierarchy.

PR #700, opened 2025-09-25
Give each model an individual data directory

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/684.
Closes https://github.com/CDCgov/pyrenew-hew/issues/699.


damonbayer:
@dylanhmorris @sbidari I will verify on azure batch before final approval.

PR #709, opened 2025-09-30
Move general datetime indexing functions from pyrenew-hew to PyRenew

cdc-mitzimorris:
This PR is the companion to https://github.com/CDCgov/PyRenew/pull/600 which adds general utilities for handing time indexing and time slicing to PyRenew by identifying code which was implemented in pyrenew-hew and is a good candidate for reuse.


tests/test_pyrenew_hew_data.py
```diff
@@ -123,15 +123,45 @@ def test_to_forecast_data(
     assert (
         (forecast_data.first_hospital_admissions_date - data.first_data_date_overall)
         / np.timedelta64(1, "D")
-    ).item() <= 6
+    ).item() <= 7
```
damonbayer: There are no changes to the `parametrize` setup, so why does this need to change?

damonbayer: >But hospital admissions are weekly data - you need at least one full week!

By this logic, we should always be jumping to the second Saturday and requiring `7 <= forecast_data.first_hospital_admissions_date - data.first_data_date_overall < 14 `, no?

PR #729, opened 2025-11-13
bump python packages

damonbayer:


damonbayer:
The failure is due to polars 1.35: https://github.com/pola-rs/polars/issues/25347

Unclear if we were relying on unintended behavior or if the change was unintentional. Going to pin 1.34 for now.

damonbayer:
Current failure has something to do with `polarbayes` and `xarray` `2025.10`. Pinning older release of `xarray`.

damonbayer:
Tests are passing here, but planning to keep open until we have a resolution on `xarray` and `polars` issues.

Issue #751, opened 2025-12-04
New `forecast_epiautogp` function


SamuelBrand1:
The steps to get to here are:

- [ ] Post-processing check (this should pass because output of `EpiAutoGP` will match that of other models.

damonbayer:
Closed in https://github.com/CDCgov/pyrenew-hew/pull/843

PR #753, opened 2025-12-04
Issue 752: Add `EpiAutoGP`

SamuelBrand1:
Following the planning in #690 , this PR adds `EpiAutoGP` to the `epiautogp` branch (which is currently just based on main) rather than `main`

This is a step towards resolving #751 (which is a substep towards #690).

### `EpiAutoGP` Structure

* Added `Project.toml` specifying package metadata, dependencies, compatibility, and external source for `NowcastAutoGP`.
* Created `src/EpiAutoGP.jl` module exporting key functions and constants along with minimal amounts of unit testing
* Added a `README.md` with`EpiAutoGP` overview
* Implemented `run.jl` as the main entrypoint script for running forecasts.

### Test data refactor

I've also added the test data refactor from #749 because its been helpful in developing at each stage, although this is out of scope so I can revert.


pipelines/tests/test_generate_test_data_lib.py
damonbayer: I think these snuck in from https://github.com/CDCgov/pyrenew-hew/pull/749

SamuelBrand1:
OK I went down the revert #749 based changes route.

PR #756, opened 2025-12-05
Issue 755: Add `epiautogp` data conversion

SamuelBrand1:
This pull request adds a new `epiautogp` submodule for `pipelines` which will house the required functions for integrating `EpiAutoGP` into the STF workflow. Initial commit adds just one library file:

* `pipelines/epiautogp/prep_epiautogp_data.py` implementing `convert_to_epiautogp_json()`, which converts surveillance data from the pyrenew-hew pipeline into the `EpiAutoGP` JSON input format, supporting both NSSP and NHSN targets, for the three main targets: Flu, Covid, RSV.

### Testing and Validation

* Added `pipelines/tests/test_prep_epiautogp_data.py` with unit tests for the conversion function. This is organised under one test class in anticipation of #749 .
* Added shell script `pipelines/tests/test_epiautogp_prep.sh` with the helper script `pipelines/tests/test_epiautogp_prep_script.py`. This is based on the existing end-to-end shell scripts, but only does test data generation and conversion to `EpiAutoGP` format.

The shell script for end-to-end testing will get updated as steps towards #690 are completed.

### Data wrangling pattern

I've followed the patterns I found in `pipelines` where I've accessed the NSSP data from `combined_training_data.tsv` (after that is generated) and NHSN data from a linked `parquet` file, even though that data is also available in `combined_training_data.tsv`. But I'm not sure why not used `combined_training_data.tsv` for both?

### DIFF
--- a/pipelines/epiautogp/prep_epiautogp_data.py
+++ b/pipelines/epiautogp/prep_epiautogp_data.py
@@ -13,6 +13,38 @@ from pathlib import Path
 import polars as pl
 
 
+def _validate_epiautogp_parameters(
+    target: str,
+    frequency: str,
+    use_percentage: bool,
+) -> None:
+    """
+    Validate EpiAutoGP conversion parameters.
+
+    The inadmissible parameter combinations are:
+    - `target` not in ['nssp', 'nhsn']
+    - `frequency` not in ['daily', 'epiweekly']
+    - `use_percentage` is `True` when target is 'nhsn' (NHSN data are always counts)
+    - `frequency` is 'daily' when target is 'nhsn' (NHSN data are only epiweekly)
+    """
+    # Validate individual parameters
+    if target not in ["nssp", "nhsn"]:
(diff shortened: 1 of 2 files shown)


SamuelBrand1:
I've made some changes, mainly around targeting the `tsv` files for data wrangling.

- Supports both daily and epiweekly data with `frequency` parameter which chooses which data file to use
- Optional percentage transformation for NSSP ED visits.

This covers some of #763 but not all of it.

pipelines/epiautogp/prep_epiautogp_data.py
```diff
@@ -0,0 +1,292 @@
...
+    """
+    if logger is None:
+        logger = logging.getLogger(__name__)
+
+    # Validate target
+    if target not in ["nssp", "nhsn"]:
+        raise ValueError(f"target must be 'nssp' or 'nhsn', got '{target}'")
+
+    # Validate frequency
+    if frequency not in ["daily", "epiweekly"]:
```
damonbayer: daily is only usable with nssp

PR #769, opened 2025-12-08
Issue 768: make `--n-chains` a pyrenew only cl arg

### LINKED ISSUE #768: Not all models need an `n-chains` argument
After #758 we still have `--n-chains` as a cli argument expected to be passed to any model using `add_common_forecast_arguments`. I think this isn't required and that only `pyrenew` needs this arg.

SamuelBrand1:
as per title.

This PR closes #768


damonbayer (changes requested):
See above.

PR #790, opened 2025-12-15
Issue 788: Add S3 dispatch for model sample processing

SamuelBrand1:
## Summary

This PR adds an extensible interface for processing location forecasts to support multiple model types and make easier to do future expansion, which unblocks #783 .

## Details

I've added a new `model_name` argument to `process_loc_forecast` function, if this is not `NA` then it lowers to `process_forecast` function and if it is `NA` (as default) then current logic is untouched.

`process_forecast` function can either be supplied a `model_type` to dispatch on a `process_model_samples` method or it detects from model name with the `detect_model_type` helper func. The functions then gathers the same data as `process_loc_forecast` and then dispatches the appropriate processing on the raw model output. 

* Added `process_model_samples` methods for `"pyrenew"` and `"timeseries"` model types, so this should work with eg `model_name = "pyrenew_e"` but I've also kept legacy interface for backwards compat.

### DIFF
--- a/hewr/tests/testthat/test_process_loc_forecast.R
+++ b/hewr/tests/testthat/test_process_loc_forecast.R
@@ -1,3 +1,20 @@
+# Common test fixtures
+fake_dir <- "/fake/dir"
+minimal_required_columns <- c("date", ".value")
+full_required_columns <- c(
+  ".chain",
+  ".iteration",
+  ".draw",
+  "date",
+  "geo_value",
+  "disease",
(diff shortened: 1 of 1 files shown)


damonbayer (approved):
I think there is probably room to make some better abstractions, but I'm in favor of unblocking https://github.com/CDCgov/pyrenew-hew/pull/783, so I have approved this PR and will work on some parallel changes in another PR.

PR #795, opened 2025-12-17
Fix to #790

SamuelBrand1:
This is another pass at #790 with full CI triggered.


SamuelBrand1:
The fix to the problem is here 

https://github.com/CDCgov/pyrenew-hew/blob/336051da7e9e75f0ac6bc9138e3baa6c61136b83/pipelines/plot_and_save_loc_forecast.R#L67-L73

The change to `hewr` meant that this needed to be changed to kwarg rather than positional args

damonbayer (approved):
Thanks @SamuelBrand1!

PR #804, opened 2025-12-19
Add `model_name` parameter for post-processing

SamuelBrand1:
This pull request changes the model selection logic for location forecast processing by introducing `model_name` parameter with auto-detection of model type. This is aimed at using the S3 method approach introduced in #790  #795 .


damonbayer (approved):
One small nit. Thanks @SamuelBrand1!

Issue #427, opened 2025-04-16
revise interface to `postprocess_forecast_batches` to not force re-processing of old forecasts


damonbayer:
\

PR #827, opened 2026-01-07
Revise `postprocess_forecast_batches` to not force re-processing of old forecasts

damonbayer:
Closes https://github.com/CDCgov/pyrenew-hew/issues/427

Also omits w models from fitting.


pipelines/azure_command_center.py
damonbayer: Need to revert some of this.

PR #838, opened 2026-01-13
Remove unused prod priors

damonbayer:
@dylanhmorris What do you think about also removing the files https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/priors/eval_priors.py and https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/priors/parameter_inference_priors.py?

### DIFF
--- a/pipelines/priors/prod_priors.py
+++ b/pipelines/priors/prod_priors.py
@@ -1,7 +1,6 @@
 import jax.numpy as jnp
 import numpyro.distributions as dist
 import pyrenew.transformation as transformation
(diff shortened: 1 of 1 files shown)


dylanhmorris:
> @dylanhmorris What do you think about also removing the files https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/priors/eval_priors.py and https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/priors/parameter_inference_priors.py?

I am in favor.

PR #874, opened 2026-02-05
Automatically set n_threads for EpiAutoGP

damonbayer:
Closes https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/859.


damonbayer:
Doesn't appear to have saved any time in CI, but I can confirm my local julia process used > 100% CPU in local testing.

PR #880, opened 2026-02-09
Remove pyrenew-hew and its tests

sbidari:
What's changed:
- Removes `pyrenew-hew` and adds `pyrenew-multisignal` as a dependency
- Removes `tests` folder corresponding to `pyrenew-hew`
- Removes pyrenew-hew test from CI run
- Updates codecov badges in README

What's not changed:
- `.github/actions/setup-pyrenew-hew` should probably be renamed to reflect current structure but pipeline works without renaming

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,14 +13,14 @@ repos:
   #####
   # Python
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.14
+    rev: v0.15.0
     hooks:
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -10,9 +10,9 @@ authors = [
   {name = "Subekshya Bidari"},
 ]
 
-
 dependencies = [
     "pyrenew",
+    "pyrenew-multisignal",
```
damonbayer: pin the current release

PR #882, opened 2026-02-10
Separate models into distinct directories

damonbayer:
## Major Changes
- [x] Model directories now contain only one `combined_data.tsv` file, which is either the epiweekly or daily data, depending on the model.
- [x] The `resolution` column is baked in to `combined_data.tsv` rather than being imputed at read time.
- [x] All `aggregated_numerator` and `aggregated_denominator` columns have been removed from outputs, since this is no longer necessary to differentiate between models.
- [x] "fusion" forecasts are now made for pyrenew numerators with both fable and epiautogp denominators
- [x] `prop_ed_visists` data is now baked in to the data files, so epiautogp doesn't need to construct the percentages itself.
- [x]  Overall there is a much greater degree of consistency and simplicity between different models, resulting in us being able to remove a huge amount code designed to work with the inconsistent data.

## Minor Changes
- [x] Some Julia formatting changes occurred from my IDE. See https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/886
- [x] `model_id` column is no longer part of the hub_quantile table, since the `model` column is now unique. See https://github.com/cdcent/cfa-hubverse-processor/issues/126
- [x] You can now fit a PyRenew E model, even if the timeseries of epiautogp outputs do not already exist. The fusion forecasts are automatically created in `forecast_pyrenew` if the timeseries of epiautogp outputs exist.
- [x] `pipelines/tests/test_end_to_end.sh` now re-installs `hewr` when it is run.
- [x] `pipelines/tests/test_end_to_end.sh` and `pipelines/tests/test_epiautogp_end_to_end.sh` run the same epiatuogp models.

## Out of scope
Timeseries script still fits multiple models at once. The resulting forecast is not problematic becasue it does not contain multiple forecasts of the same signal. We could consider modifying this at a later time.

### DIFF
--- a/dagster_defs.py
+++ b/dagster_defs.py
@@ -61,7 +61,7 @@ user = os.getenv("DAGSTER_USER")
 workdir = "cfa-stf-routine-forecasting"
 local_workdir = Path(__file__).parent.resolve()
 
-tag = "dmb_timeseries_dir_rework"
+tag = "latest"
 image = f"ghcr.io/cdcgov/cfa-stf-routine-forecasting:{tag}"
 
 default_config = ExecutionConfig(


pipelines/epiautogp/create_samples_from_epiautogp_fit_dir.R
dylanhmorris: Perhaps flag for the reader that the point of this script is as a workaround for Julia parquet writer not handling dates?

pipelines/fable/forecast_timeseries.py
```diff
@@ -6,22 +6,34 @@
...
+        args.append("--overwrite-daily")
+
+    run_r_script(
+        "pipelines/data/generate_epiweekly_data.R",
+        args,
+        function_name="generate_epiweekly_data",
+    )
+    return None
+
+
```
dylanhmorris: As noted above, why here rather than common_utils?

pipelines/utils/create_prop_samples.R
```diff
@@ -0,0 +1,223 @@
...
+library(argparser)
+library(dplyr)
+library(tidyr)
+library(stringr)
+
+augment_samples_with_obs <- function(samples, obs) {
+  first_forecast_date <- min(samples$date)
+  sample_resolution <- unique(samples$resolution)
+
+  checkmate::assert_scalar(sample_resolution)
```
dylanhmorris: Question: should we not force the obs to know is resolution too and confirm compatibility?

dylanhmorris (approved):
A few minor questions and suggestions, @damonbayer. Happy for you to merge without re-review.

pipelines/fable/forecast_timeseries.py
damonbayer: [`1b5d009` (this PR)](https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/882/commits/1b5d0094c6e28d7b41fc71b42984f3e1c4bf812c)

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -15,16 +15,28 @@
 import polars as pl
 
 from pipelines.data.prep_data import process_and_save_loc_data
-from pipelines.pyrenew_hew.forecast_pyrenew import generate_epiweekly_data
+from pipelines.fable.forecast_timeseries import generate_epiweekly_data
```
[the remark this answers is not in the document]
damonbayer: [`1b5d009` (this PR)](https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/882/commits/1b5d0094c6e28d7b41fc71b42984f3e1c4bf812c)

pipelines/utils/create_prop_samples.R
damonbayer: [`f4f1962` (this PR)](https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/882/commits/f4f19626df9c090d1797bd76e9b0e68b7c80d02e)

PR #888, opened 2026-02-13
Containerfile refactor for security, cache performance, and clarity.

jkislin:
In lieu of review comments on #714, I want to split this into its own PR for review. 
If approved, the dagster version of this will just entail a single extra `COPY ./dagster_defs.py .` line in my main PR.

**Motivations:**
1. `COPY . .` is a security risk
     - Images can contain credential files or other untracked artifacts if contributors are not careful.
2.  Cache performance
     - Since docker's cache is linear, it's good practice to put expensive operations earlier. 
     - For example, installing the EpiAutoGP julia package and it's dependencies takes 2 or more minutes to build, but more actively developed python files are copied in earlier, breaking the build cache frequently when they change.
3. Organization/clarity
     - Steps are now thematically related separated by comment sections - feel free to push back if you find this heavy handed.

**Summary of changes**:
1. The `Containerfile` is reorganized according to the points above.
2. The project is called `cfa-stf-routine-forecasting` in the `pyproject.toml` now - this is also accordingly the name of the uv virtual environment. The package built by uv with the `pyproject.toml` remains `pipelines`.
3. I've reimplemented cacheing in Github Actions in the `containers.yaml` workflow.
4. Updated the `Makefile` to build containers based on your branch name - same as the above-mentioned `containers.yaml` workflow. Also added `container_explore` to quickly jump into bash within the container you built.
5. `dagster_defs.py` now references the container based on your branch name, also mirroring the functionality above.
6. Given our new understanding that Azure Batch pools' autoscale formulae can be "requeued" (thanks @natemcintosh), I have defaulted all runs to Azure Batch, even dev runs. This will also help us not soft-brick our VMs :) 

In totality, these changes should hasten and simplify local prototyping.

Thanks for your consideration! I welcome feedback.


pyproject.toml
```diff
@@ -30,6 +30,9 @@ dependencies = [
     "jax<0.8.2",
 ]
 
+[tool.hatch.build.targets.wheel]
+packages = ["./"]
```
jkislin: This was suggested by a uv error message. I'm investigating if there's a better practice.

pyproject.toml
```diff
@@ -1,9 +1,9 @@
 [project]
-name = "pipelines"
+name = "cfa-stf-routine-forecasting"
 version = "0.1.0"
 description = "Pipelines for forecasting respiratory diseases"
 readme = "README.md"
-requires-python = ">=3.13"
+requires-python = ">3.13, <3.14"
```
damonbayer:
```suggestion
requires-python = ">=3.13, <3.14"
```

dagster_defs.py
```diff
@@ -61,7 +62,14 @@
...
 
-tag = "latest"
+# If the tag is prod, use 'latest'.
+# Else iteratively test on our dev images
+# (You can always manually specify an override in the GUI)
+repo = Repository(os.getcwd())
+current_branch_name = repo.head.shorthand
+tag = (
+    "latest" if is_production or current_branch_name == "main" else current_branch_name
+)
```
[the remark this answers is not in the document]
jkislin: I've taken the approach that forecast_pyrenew() takes where it will check .git first (through pygit2) but will fallback on the build arg $GIT_BRANCH_NAME if it doesn't find it, and `unknown` as a final fallback.

jkislin:
@damonbayer tests are passing, containers are building, dagster is running!

Let me know if you think any other changes are needed!

PR #902, opened 2026-02-20
Add new timeseries variants to dagster

jkislin:
WIP! @damonbayer. 

I've started with a commit that will default the dagster config to use a given branch's dev version of the container; it can always be overridden. Note that before you can test an image with dagster, it of course has to be built first.

Targeting/based on your branch, but we can swap to main if you merge first.

### DIFF
--- a/dagster_defs.py
+++ b/dagster_defs.py
@@ -328,9 +328,7 @@ def daily_timeseries_e(context: dg.AssetExecutionContext, config: ModelConfig):
 
 
 # Weekly Timeseries E
-@dg.asset(
-    partitions_def=pyrenew_multi_partition_def,
-)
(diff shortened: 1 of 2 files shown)


dagster_defs.py
```diff
@@ -61,7 +61,16 @@
 workdir = "cfa-stf-routine-forecasting"
 local_workdir = Path(__file__).parent.resolve()
 
-tag = "latest"
+# If the tag is prod, use 'latest'.
+# Else iteratively test on our dev images
+tag = (
```
damonbayer: We already import `pygit2` to get the branch name for logging

https://github.com/CDCgov/cfa-stf-routine-forecasting/blob/6e172a6faa1de287c83bd1f2d6d9aa33302a3dd8/pipelines/pyrenew_hew/forecast_pyrenew.py#L45-L46

PR #911, opened 2026-02-26
updating cfa-dagster to fix wsl dev

giomrella:
This changes the default Dagster port to 4000 to work with WSL:
image


damonbayer:
Verified to work on my Windows VAP. Thanks @giomrella!

PR #946, opened 2026-03-10
Dynamic graph assets, date partitions, and declarative automation

jkislin:
@damonbayer @giomrella - this is still in development, but has major ramifications for future development, so I think a review is warranted even as I continue to test.

### This update makes significant changes to the way we think about our assets:
1. Dynamic Graph Assets - gone are partitions for `disease` and `location`; instead, these are "graph dimensions" which scale out compute based on values we supply in config. This allows us to use date as a partition, which opens up [Declarative Automation](https://docs.dagster.io/guides/automate/declarative-automation).
2. Declarative Automation - we can specify when we want specific assets to materialize, or on what events (and in what ways) we'd like them to. We're only just scratching the surface here. Scheduling asset backfills is no longer necessary, and as such I've removed a significant amount of job and schedule code.
3. Data Availability Checks as Upstream Assets - these are functionally equivalent to external data sensors, but as we plan to bring upstream data into dagster, writing the sensors as assets makes for less rewriting later.
4. Asset Groups - we can apply automation conditions to thematically grouped assets; assets are grouped as we define in the UI. For example, I've created three groups: `WeeklyForecast`, `WeeklyForecastArchive`, and `ExperimentalEpiAutoGP`.

### Changes to UX/UI:
1. We will (potentially) no longer use Azure Container App Jobs to launch jobs for Azure Batch Execution. 
2. The Backfill UI is no more; the timeline view for a backfill is no more; runs will dynamically spawn the timeline view within their own UI. 
3. Dagster now keeps track as "Date" as a partition, so you won't modify it in config, you'll modify it in the partition selector. 

### Future development enabled by this PR
- Once we get this right, we can look at the blob folder structure of our outputs and directly materialize outputs rather than maintaining dummy string outputs to track success. This has ramifications for I/O management.
- NHSN, NSSP, NWSS, and other data can be brought to dagster and tracked as dependencies here and in other projects across code locations.

Images and discussion to come.

### DIFF
--- a/dagster_defs.py
+++ b/dagster_defs.py
@@ -3,6 +3,8 @@ import datetime as dt
 import os
 from pathlib import Path
 
+# Direct use of dagster
+import dagster as dg
 import requests
 from azure.identity import DefaultAzureCredential
 from azure.storage.blob import BlobServiceClient
@@ -28,9 +30,6 @@ from forecasttools import location_table
 from pygit2.repository import Repository
 from pyrenew_multisignal.hew.utils import flags_from_hew_letters
 from pytz import timezone
-
-# Direct use of dagster
-import dagster as dg
 
 # Local constant imports
 from pipelines.batch.common_batch_utils import (


dagster_defs.py
```diff
@@ -277,53 +284,162 @@ class PostProcessConfig(CommonConfig):
...
+
+    # Disease and Locations are our "Graph Dimensions".
+    disease = context.graph_dimension["diseases"]
+    location = context.graph_dimension["locations"]
+
+    # Date is the daily partition we use
+    date = context.partition_key
+
+    # This is the critical piece that allows our graph asset strategy
+    context.register_output(
```
giomrella: You know I put a lot of work into making sure `context.register_output()` could only be called as the first line of the `@dynamic_graph_asset` decorator so users could only access the context, config, and dependent asset variables that were in scope, but you found a legitimate workaround by passing the context into a new scope. If this works, I think its OK.

jkislin:
> > 4. Asset Groups - we can apply automation conditions to thematically grouped assets; assets are grouped as we define in the UI. For example, I've created three groups: `WeeklyForecast`, `WeeklyForecastArchive`, and `ExperimentalEpiAutoGP`.
> 
> What are the use cases for each of these groups?

So we can define scheduling and execution config at a level of granularity in between "just this asset" and "all of the assets".

> > 3. Dagster now keeps track as "Date" as a partition, so you won't modify it in config, you'll modify it in the partition selector.
> 
> Can you say more about this choice? I understand why the locations and diseases are no longer partitions, but why is the date now a partition?

We went into it on Teams, but two reasons originally :
1. Asset materialization granularity opens up more features for materialization and lineage tracking. I.e. instead of returning "pyrenew_e_output" we can point to the pyrenew e outputs in 2026-05-03_forecasts in blob. 
2. You can re-execute runs from a previous date - we discussed that this is not something we'd want to do, or if we did, we wouldn't want to call it the same model output. 

After our conversation, I'm not convinced either of these are particularly necessary, at least at this point in time.

I have some ideas that I'll try next week. I think we don't need the date partitions on the model assets, but having date partitions on the upstream data check assets (i.e. did NHSN post yet for _today_?) combined with declarative automation would give us the flexible scheduling abilities we're looking for without unnecessarily complicating the model asset materializations.

**In summary - still experimenting; PR should not be merged as is!**

jkislin:
Note:
This has successfully launched on my local machine as intended, submitting to Azure Batch, with outputs to `test-output`.

PR #965, opened 2026-03-17
Change `loc_abbr` arg to `loc_abb` in get_nhsn_hrd()

sbidari:


sbidari:
> Don't see a problem with this but don't know the motivation. Can you explain briefly?

This codebase [seems](https://github.com/search?q=repo%3ACDCgov%2Fcfa-stf-routine-forecasting%20loc_abb&type=code) to use `loc_abb` elsewhere. 
So I changed `loc_abbr` used in backtest codebase to `loc_abb` in `cfa.stf.data` but missed one function `get_nhsn_hrd()`

PR #1008, opened 2026-04-09
Removed configurability of output subdirectory in dagster; get forecast date directly from daily partition key everywhere

jkislin:
Minor update to make sure dates resolve the same everywhere:
1. We no longer allow the date to be switched in the config value on the launchpad. 
2. Instead of computing today's date with a function, we just let dagster do it with its built in Daily Partition definitions and reference that.

PR #1009, opened 2026-04-09
Use context.log.info instead of context.log.debug

jkislin:
The context log `debug` level is hidden by default. I propose changing our default pattern to use `context.log.info()` over `context.log.debug()` unless we can think of something we'd only want an engineer to see versus a user. 

image

`context.log.warning` and `context.log.error` are also useful - and we use them here too (though I haven't added or changed anything wrt those in this PR)


jkislin:
Wait for merge on #1008 before marking this ready.

damonbayer (approved):
@jkislin This looks good. Feel free to merge after resolving the conflict.

PR #1011, opened 2026-04-10
Create dagster assets for fusion models

### LINKED ISSUE #895: Clarify PyRenew / Fable / EpiAutoGP dependency relationship
After https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/882 lands, we no longer need PyRenew to depend on Timeseries or EpiAutoGP. Instead, we can create fusion models any time PyRenew, Timeseries, or EpiautoGP models are fit.

damonbayer:
Closes https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/895

This PR moves our "fusion" forecasts (which combine two upstream forecasts into a new forecast) to their own assets.

Previously, model fusion was performed inside of `forecast_pyrenew`, resulting the `pyrenew-*E*` models having a confusing dependency on the timeseries models.

Now the dependency graph accurately reflects the dependency between models.

CleanShot 2026-04-15 at 15 36 52@2x


jkislin:
I like this a lot. I will dive in.

damonbayer:
I should also rework the end to end test and github tests to mimic this behavior.

dagster_defs.py
```diff
@@ -219,6 +225,10 @@ class TimeseriesConfig(ModelBaseConfig):
     n_samples: int = 400 if not is_production else 2000  # Total samples for timeseries
 
 
+class FusionConfig(ModelBaseConfig):
```
jkislin: It might be better practice to avoid defining a new class until we have unique configuration entries for it.

dagster_defs.py
```diff
@@ -473,23 +484,111 @@ def _run_pyrenew_model(
     )
 
 
+def get_model_loc_dir(
```
jkislin, line 483: These three functions - I'm inclined to want them consolidated into the final function that's the only one of the three invoked more than once - `_fuse_pyrenew_timeseries()` .

(Unless there's a clear roadmap where we'd expand out into more fusion models and other models where you'd need to find the location of an existing model output.)

- `get_model_loc_dir()` is called only once outside its definition.
- `_run_fusion_model()` is called only once outside its definition.

Pre-emptive abstraction is something we do at CFA a lot, and Gio and I are increasingly of the opinion that it's a quiet form of technical debt.

dagster_defs.py
```diff
@@ -473,23 +484,111 @@ def _run_pyrenew_model(
...
+    )
+
+
 # ---------- Automation Conditions ----------
 # A condition that will run as soon as all dependencies are available, but only
 # if the asset has not been materialized yet for that day
 eager_once = dg.AutomationCondition.eager() & dg.AutomationCondition.missing()
 
-# ---------- Pyrenew Assets ----------
+weekly_forecast_initial = {
```
jkislin: Similarly, repeating ourselves a little bit here might be more transparent to future users than variablizing our decorator args for parameter expansion (`**`). "Never repeat yourself" vs "be clear" are at odds here.

jkislin:
```suggestion
weekly_forecast_initial_args = {
```
I'm somehow more ok with this with the inclusion of 'args' in the name. Maybe pedantic, but it adds to clarity!

dagster_defs.py
```diff
@@ -602,14 +683,99 @@ def pyrenew_hew(
     _run_pyrenew_model(context, config, "hew")
 
 
+# ---------- Fusion Forecasts ----------
+
+weekly_forecast_fusion = {
```
jkislin:
```suggestion
weekly_forecast_fusion_args = {
```
Ditto to above.

jkislin (approved):
@damonbayer I think this is a great update. I have some opinions as to the level of abstraction in the new functions and parameter expansion, maybe you have some thoughts in response. Otherwise LGTM!

dagster_defs.py
damonbayer: I can go with this. The idea is that this provides the arguments for the models that are not dependent on any other models, so you would want a way to change all of those arguments at the same time.

dagster_defs.py
damonbayer: This is a good comment, but I know that these will be used again when we add the epiautogp model.

dagster_defs.py
damonbayer: [`ef0bf0f` (this PR)](https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1011/commits/ef0bf0ff27f39adefc9ab7a851ddb28e30a04639)

dagster_defs.py
```diff
@@ -155,7 +161,7 @@
                     # f"/home/{user}/.azure:/root/.azure",
                     # bind current file so we don't have to rebuild
                     # the container image for workflow changes
-                    # blob container mounts for cfa-stf-routine-forecasting
+                    # get_ blob container mounts for cfa-stf-routine-forecasting
```
damonbayer:
```suggestion
                    # blob container mounts for cfa-stf-routine-forecasting
```

PR #1012, opened 2026-04-10
Make post procesing artifact available

damonbayer:
and bump some actions


damonbayer:
Artifact looks as expected, though we had no way to know that, until now!

PR #1016, opened 2026-04-11
Prune daily partitions at initialization

damonbayer:
This will allow us to have a much shorter list of partitions, since we only ever want to run the current day.

### DIFF
--- a/dagster_defs.py
+++ b/dagster_defs.py
@@ -2,6 +2,7 @@
 import datetime as dt
 import os
 from pathlib import Path
+from zoneinfo import ZoneInfo
 
 # Direct use of dagster
 import dagster as dg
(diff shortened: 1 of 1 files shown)


giomrella (approved):
From our conversation in Teams, the partition definition will be evaluated at import-time, so I would expect it to show only the latest partition the first day this code is deployed with the partition list growing each day the code location is online. 

It would probably be a good idea to deploy this to the production server overnight to confirm this behavior

damonbayer:
> It would probably be a good idea to deploy this to the production server overnight to confirm this behavior

Done. Will check on this tomorrow.

giomrella:
As expected, there is a new partition for today and yesterday's partition is still here:
image

PR #1024, opened 2026-04-16
Add julia formatter to pre-commit

### LINKED ISSUE #886: Add julia formatter to CI
My editor is making formatting changes [here](https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/882/).
I have seen @SamuelBrand1 recommend [runic](https://github.com/fredrikekre/Runic.jl), though EpiAutoGP doesn't appear to have used it.

It can be added to pre-commit with
```yaml
  # Julia
  - repo: https://github.com/fredrikekre/runic-pre-commit
    rev: v2.0.1
    hooks:
      - id: runic
```

damonbayer:
Closes https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/886

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,7 @@
+ci:
+  skip:
+    - runic
+
 repos:
   # Basic file cleanliness
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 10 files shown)


damonbayer:
Seems like none of the julia formatters work in pre-commit.ci because they require julia to be installed (see, for example, https://github.com/domluna/JuliaFormatter.jl/issues/800). I added the hook for local use only.  cc @SamuelBrand1

PR #1031, opened 2026-04-17
cfa-dataops for upstream data; flexible eager/cron automation for initial forecasts

### LINKED ISSUE #1032: Trigger weekly forecast workflow with NHSN and NWSS dataops assets

jkislin:
#### Upstream Data
- NHSN and NWSS should be referenced from `cfa-dataops-scheduler` now that those assets are available on the Dagster production server.
    - When not in production, we use `AssetSpec` assets as placeholders to model our entire workflow. These are removed and replaced with the external co-located assets in the prod server from `cfa-dataops-scheduler`. 

#### Automation    
- We eagerly materialize initial forecast models on Wednesdays during Working Hours (6am-8pm - liberal definition)
- We eagerly materialize fusion forecast models and postprocessing.

#### Misc
- Moved `hewr` installation to below Julia package management as the former is both more brittle than the latter and takes less time.


dagster_defs.py
```diff
@@ -520,22 +458,57 @@ def _fuse_pyrenew_timeseries(
...
+# without needing to actually colocate the assets. We only do this on the dev server
+
+if not is_production:
+    nssp_gold_v1 = dg.AssetSpec(
+        "nssp_gold_v1", partitions_def=daily_partitions_def, group_name="Upstream"
+    ).with_io_manager_key("prod_io_manager")
+
+    nhsn_hrd = dg.AssetSpec(
+        "nhsn_hrd", partitions_def=daily_partitions_def, group_name="Upstream"
+    ).with_io_manager_key("prod_io_manager")
```
giomrella: This won't cause a problem because we're not using the IOManagers to pull asset data, but the `nssp_gold_v1` asset uses the `ADLS2FilesystemIOManager` and the `nhsn_hrd` asset doesn't use any IOManager.

dagster_defs.py
```diff
@@ -159,7 +158,7 @@
                     # f"/home/{user}/.azure:/root/.azure",
                     # bind current file so we don't have to rebuild
                     # the container image for workflow changes
-                    # blob container mounts for cfa-stf-routine-forecasting
+                    # get_ blob container mounts for cfa-stf-routine-forecasting
```
damonbayer:
```suggestion
                    # blob container mounts for cfa-stf-routine-forecasting
```

dagster_defs.py
```diff
@@ -520,22 +458,58 @@ def _fuse_pyrenew_timeseries(
...
+# while in development. They do not materialize.
+# They are replaced with true assets in production where
+# other code locations are able to be referenced.
+
+if not is_production:
+    nssp_gold_v1 = dg.AssetSpec(
+        "nssp_gold_v1", partitions_def=daily_partitions_def, group_name="Upstream"
+    ).with_io_manager_key("prod_io_manager")
+
+    nhsn_hrd = dg.AssetSpec(
```
damonbayer: I think we want these all to be `nhsn_hrd_prelim`, which is the dataset that is updated on Wednesdays. `nhsn_hrd` is updated on Fridays.

dagster_defs.py
```diff
@@ -520,22 +458,58 @@ def _fuse_pyrenew_timeseries(
...
+weekly_forecast_initial_asset_args = {
     "partitions_def": daily_partitions_def,
     "graph_dimensions": ["diseases", "locations"],
-    "automation_condition": dg.AutomationCondition.on_cron(
-        cron_schedule="0 0 * * WED", cron_timezone="America/New_York"
-    ),
     "group_name": "WeeklyForecast",
+    "automation_condition": (
+        dg.AutomationCondition.eager()
+        & dg.AutomationCondition.cron_tick_passed(
```
damonbayer: Will the switch to `cron_tick_passed` resolve the issue @giomrella mentions here?

https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1039#issuecomment-4299555558

uv.lock
damonbayer, line 1: Doesn't seem like this should change at all. Revert it or upgrade your uv and do a fresh lock/sync?

damonbayer (changes requested):
I think we need to change `nhsn_hrd` -> `nhsn_hrd_prelim`.

damonbayer (approved):
Thanks @jkislin! Please address or reply to https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1031#discussion_r3133745945 before merging.

dagster_defs.py
jkislin: Removed.

dagster_defs.py
jkislin: @giomrella any proposals on how we would test this in dev?

jkislin: This has been superseded by a custom condition `IsWeekday()`

jkislin:
@giomrella and @damonbayer, tested and ready for re-review.

dagster_defs.py
```diff
@@ -733,15 +724,12 @@ def postprocess_forecasts(
...
-    jobs=collected_defs["jobs"],
-    sensors=collected_defs["sensors"],
-    schedules=collected_defs["schedules"],
+    **collected_defs,
     resources={
-        # This IOManager lets Dagster serialize asset outputs and store them
+        # These IOManagers let Dagster serialize asset outputs and store them
         # in Azure to pass between assets
         "io_manager": ADLS2PickleIOManager(),
+        "prod_io_manager": ADLS2PickleIOManager(use_production=True),
```
giomrella: I don't see this used

dagster_defs.py
```diff
@@ -750,8 +738,8 @@ def postprocess_forecasts(
     },
     # You can put a comment after azure_batch_config to solely execute with Azure batch
     executor=dynamic_executor(
-        default_config=azure_batch_execution_config,
-        # default_config=basic_execution_config,
+        # default_config=azure_batch_execution_config,
+        default_config=basic_execution_config,
```
giomrella: This was probably for testing

dagster_defs.py
```diff
@@ -520,22 +455,108 @@ def _fuse_pyrenew_timeseries(
...
     "partitions_def": daily_partitions_def,
     "graph_dimensions": ["diseases", "locations"],
-    "automation_condition": dg.AutomationCondition.on_cron(
-        cron_schedule="0 0 * * WED", cron_timezone="America/New_York"
-    ),
     "group_name": "WeeklyForecast",
+    "automation_condition": (
+        # We specifically don't want these to run unless it's Wednesday
+        # 0=monday,1=tuseday,2=wednesday,etc.
+        IsWeekday(2) & dg.AutomationCondition.eager()
```
giomrella: Would this produce fewer evaluations if the order were reversed? e.g. `dg.AutomationCondition.eager() & IsWeekday(2)`

jkislin: If it remains like this:
1. Will evaluate the first operand as true every tick on Wednesday.
2. Will evaluate the first operand (and thus both, given programming convention) as false every tick not on Wednesday.

If we reverse the order, `eager()` will be evaluated every tick, which would make Wednesday have less compute overhead for evaluation, but it would also cause evaluation to happen anytime upstream updates even when not on Wednesday.

As for actual execution - the order shouldn't matter. It's just a sensor evaluation overhead thing.

Pick our poison? What do you think?

giomrella: I'd prefer whichever option doesn't produce an asset evaluation every tick (30 seconds). If they both do, then it's fine as-is.

jkislin: I'll put `eager()` first, since the upstream assets only go once per day at most.

dagster_defs.py
jkislin: Removed

PR #1041, opened 2026-04-23
Add eager condition to fusion models; remove all nwss/w entities

jkislin:
- Fusion models should launch as soon as their upstream dependencies are available, and then re-launch when any updates appear upstream.
- We are not using `w` models currently. Let's remove them and add them back if eventually needed.

PR #1042, opened 2026-04-23
exclude wyoming from fusion models

### LINKED ISSUE #1040: Fusion models should filter out WY

jkislin:
New class `FusionConfig` which is identical to `ModelBaseConfig` except it filters out `WY` from the config, similar to `PyrenewEConfig`

Fixes #1040


jkislin:
@damonbayer merging without waiting for review; separate PR for paper-trail.

This fixes errors got when testing #1041 and closes #1040

PR #1061, opened 2026-05-12
change Parquet for DuckDB in EpiAutoGP

### LINKED ISSUE #898: Explore alternative parquet exporters from julia
All `pipelines/epiautogp/create_samples_from_epiautogp_fit_dir.R` does is convert the data column to a `date` type because this isn't supported in https://github.com/JuliaIO/Parquet.jl. We may be able to use https://gitlab.com/ExpandingMan/Parquet2.jl or https://github.com/duckdb/duckdb instead to omit the need for this step.

SamuelBrand1:
This PR closes #898 by swapping out `Parquet.jl` for `DuckDB.jl` (the julia frontend to duckdb) in the `EpiAutoGP` package.

Also changed:

- Removed the R step to do conversion for strings to date types.
- Added an interop test that uses EpiAutoGP forecast output to parquet, and then reads them back with 1) `polars` and checks that dates are correct and understood as `pl.Date` type and 2) `forecasttools:read_tabular` and does the same checking except for the check is `inherits(samples$data, "Date")`. ~**NB: this test is skipped if julia in present to generate the fixture. Therefore, it will likely get skipped in R/py testing. This is to avoid latency to check this by installing julia every CI check.**~
- pinned the unit tests to use `julia1.11` to match the integration tests. 

EDIT: I've noted that julia is installed for testing in `.github/test.yaml`. So the strategy is to move the "use EpiAutoGP to generate forecasts and check it works from R/py" test to the end and it will use the already pre-compiled EpiAutoGP env.


damonbayer (approved):

damonbayer:
Thanks @SamuelBrand1! Grateful to have you back working on Epiautogp 🙂

PR #1067, opened 2026-05-13
Fix issue 1062: update interop tests

### LINKED ISSUE #1062: Investigate postprocessing CI failure
http://github.com/CDCgov/cfa-stf-routine-forecasting/actions/runs/25741997237/job/75631275869#step:6:81

I suspect I was too eager in merging https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1061/

Edit: yes, that did eventually fail https://github.com/CDCgov/cfa-stf-routine-forecasting/actions/runs/25737684926/job/75635485396?pr=1061

SamuelBrand1:
This PR addresses the post-processing failure found in #1062 to close.

The problem was cross-model output schema drift where `EpiAutoGP` `output_type_id` was f64 and the eventual combine step was expecting i32. 

```python
cast_options=pl.ScanCastOptions(integer_cast="allow-float")
```

would allow this (I think) for one set of schema but the mismatch causes the error. 

The fix was to coerce `EpiAutoGP` into outputting to the common type rather than doing normalisation later on (I don't want to fiddle with the prod code).

I also beefed up the unit test that would have caught this by generating fixture output from each model and running the combination post-processing function on them so this should catch schema drift in the future.

### DIFF
--- a/EpiAutoGP/src/output.jl
+++ b/EpiAutoGP/src/output.jl
@@ -110,7 +110,7 @@ function create_forecast_df(results::NamedTuple, output_type::PipelineOutput)
         DataFrame(
             :date => forecast_dates,
             Symbol(".value") => sampled_values,
-            Symbol(".draw") => fill(draw, length(sampled_values))
+            Symbol(".draw") => fill(Int32(draw), length(sampled_values))
         )
     end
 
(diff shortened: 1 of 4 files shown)


damonbayer (approved):

damonbayer:
Thanks @SamuelBrand1!

PR #1068, opened 2026-05-13
DRAFT: run EpiAutoGP direct from `juliacall`

### LINKED ISSUE #897: Rewrite epiautogp as a python wrapper around nowcastautogp

SamuelBrand1:
This is a draft/WIP PR for trying a maximalist approach to resolving #897 .

In this design, we can remove all the `.jl` files and rely on https://github.com/JuliaPy/pyjuliapkg for julia deps and run a minimal version of model directly using `juliacall`.

This means that the input/output is handled by python, and the actually forecasting code reduces the the string `JULIA_HELPER_CODE`. This could be a cleaner way of interfacing to `NowcastAutoGP` assuming that prod bugs don't surface in implementation checks.

NB: I haven't removed all the julia files in this draft _yet_. This is really a design example for @sbidari @dylanhmorris @damonbayer to consider.

Note that only surfaced pain point was some incompat with the `dev` dep group (see `pyproject.toml`)

EDIT: there is a second surfaced pain point about config on multithread jointly with python and julia.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,10 +28,16 @@ dependencies = [
 
 [tool.hatch.build.targets.wheel]
 packages = ["pipelines", "cfa"]
-include = ["cfa/stf/forecasttools/test_data/*.nc"]
+include = [
+  "cfa/stf/forecasttools/test_data/*.nc",
+  "pipelines/juliapkg.json",
+]
 
 [tool.hatch.build.targets.sdist]
-include = ["cfa/stf/forecasttools/test_data/*.nc"]
+include = [
+  "cfa/stf/forecasttools/test_data/*.nc",
+  "pipelines/juliapkg.json",
+]
 
 [build-system]
 requires = ["hatchling"]
@@ -52,6 +58,17 @@ test = [
     "pytest>=8.3.2",
     "pytest-cov>=5.0.0",
     "pytest-mpl>=0.17.0"
+]
+epiautogp-julia = [
+    "juliacall>=0.9.31",
+]
+
+[tool.uv]
+conflicts = [
+    [
+      {group = "dev"},
+      {group = "epiautogp-julia"},
+    ],
 ]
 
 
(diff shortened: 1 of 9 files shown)


SamuelBrand1:
Hmmmm multithreading sounds risky https://juliapy.github.io/PythonCall.jl/stable/juliacall/#py-multi-threading.

I'll leave that here, but I think this approach might be nice to look at but feels too risky. AutoGP relies heavily on multi-threaded.

pyproject.toml
```diff
@@ -53,6 +59,17 @@ test = [
...
+    "juliacall>=0.9.31",
+]
+
+[tool.uv]
+conflicts = [
+    [
+      {group = "dev"},
+      {group = "epiautogp-julia"},
+    ],
+]
```
damonbayer, line 72: What is the conflict? You would certainly want to be able to have all of these deps loaded for local development.

SamuelBrand1: `cfa-dagster[dev]` down the dep chain requires `dagster-dg-core` which requires `tomlkit<0.13.3` whereas the juliacall in its dep chain has `juliapkg>=0.1.21` which wants `tomlkit>=0.13.3`

SamuelBrand1: https://github.com/dagster-io/dagster/blob/6ff640471fa43fda9c2d0e38c00bc0fe7714e398/python_modules/libraries/dagster-dg-core/pyproject.toml#L21

SamuelBrand1: https://github.com/JuliaPy/pyjuliapkg/blob/0aa8af1049b5bcec99710da305bd029fb0b2b0a8/pyproject.toml#L9

SamuelBrand1: Although I'd quite like to know more about the "bug in this version" comment!

damonbayer: https://github.com/dagster-io/dagster/pull/30513

SamuelBrand1: But then what was the bug? I can't see their buildkite.

damonbayer:
Closing in favor of https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1070

PR #1075, opened 2026-05-14
Make it easier to test on the prod server

jkislin:
This is controversial. Let me know if we want just one option or the other (I enabled this in both the Makefile and Github Actions) or neither.

### DIFF
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -77,7 +77,7 @@ jobs:
 
     deploy_dagster_code:
       needs: build-pipeline-image
-      if: github.ref == 'refs/heads/main' || ${{ github.event.inputs.push_to_dagster == "yes" }}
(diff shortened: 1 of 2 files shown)


damonbayer:
@jkislin Can you explain the controversy? This seems nice!

jkislin:
> @jkislin Can you explain the controversy? This seems nice!

I think the controversy is in having the ability to push to prod server in the `Makefile` - it's arguably a bit risky to have this exposed so ergonomically on the cli. That said I'd find it personally very nice to have, and nothing's stopping someone from saving a code snippet to do the same thing.

The github actions workflow dispatch functionality I think is also useful and a little less controversial - at least you need to click to confirm there.

damonbayer (approved):
I dig it!

giomrella:
I think this is a good idea, though I have controversial thoughts about putting jobs in Makefile vs. Dagster defs..

giomrella (approved):

Issue #1058, opened 2026-05-11
NHSN nowcast ingestion


SamuelBrand1:
NSHN nowcasts are available on blob in hubverse format.

The transformation step here is taking that format and recreating the vector of vector nowcasts that EpiAutoGP works with.

PR #1082, opened 2026-05-19
Issue 1057: Nowcast abstraction and use in NSSP nowcasting for `epiautogp`

### LINKED ISSUE #1057: NSSP nowcast ingestion
The path for NSSP nowcast ingestion is quite well set in `pyrenew` context. We have nowcast CDF generated for each location. This can be used to construct a point estimate of "eventual" data.

This single point estimate can be ingested as a `n=1` batch nowcast into EpiAutoGP.

SamuelBrand1:
This pull request introduces a nowcasting abstraction for `pipelines` models aimed at use by `epiautogp` when doing daily NSSP target forecasts in the _first_ instance, but flexible enough to get extended for other nowcasting approaches/frequencies/targets.

The basic ideas for the nowcasting are:

- `NowcastData`: currently just a list of dates and a list of lists of nowcasts, following the default vector-of-vector approach in `NowcastAutoGP`. Not currently extended to other stratifications except dates.
- `NowcastSource` protocol which defines an interface for `get_nowcast_data` method that has to return a `NowcastData` obj.
- Concrete implementations of the protocol:
  - `FixedNowcast`... this will probably get replaced in resolving #1058 
  - `NsspRightTruncationNowcast`, this implements the standard inflation estimator using a bit of nuance around low values, see https://baselinenowcast.epinowcast.org/articles/model_definition.html#zero-handling-approximation for reference implementation. This is a bit more generic that "NSSP" but I've left NSSP in the name because its pretty hard linked to using the NNH right truncation PMF estimates in the reality of this pipeline.

**Pipeline and API enhancements:**

* Updated the `ForecastPipelineContext` dataclass to include a `nowcast_source` member, and modified `setup_forecast_pipeline` to resolve and attach the nowcast source based on user input and available data.
* Extended the main forecast entry point and CLI (`forecast_epiautogp.py`) to accept `param_data_dir`, `nowcast_source_name`, and `right_truncation_pmf` as parameters and command-line arguments. These get resolved to a particular nowcasting approach.

**Data preparation and JSON output:**

* Refactored the data preparation and JSON conversion logic to generate nowcast data using the resolved nowcast source, and included the resulting nowcast dates and reports in the`epiautogp` input JSON. 

**Imports and public API:**

* Updated module imports and `__all__` in `__init__.py` to expose `NowcastData` and `NowcastSource` for use throughout the pipeline.

### DIFF
--- a/pipelines/epiautogp/epiautogp_forecast_utils.py
+++ b/pipelines/epiautogp/epiautogp_forecast_utils.py
@@ -15,8 +15,8 @@ from typing import Any, Literal
 import polars as pl
 
 from pipelines.data.prep_data import get_pmfs, process_and_save_loc_data
-from pipelines.epiautogp.nssp_nowcast import NsspRightTruncationNowcast
 from pipelines.epiautogp.nowcast import NowcastSource
+from pipelines.epiautogp.nssp_nowcast import NsspRightTruncationNowcast
 from pipelines.utils.common_utils import (
     append_prop_data_to_combined_data,
     calculate_training_dates,
@@ -157,6 +157,7 @@ class ForecastPipelineContext:
         model_fit_dir_to_hub_tbl(model_fit_dir)
         self.logger.info("Postprocessing complete.")
 
+
 def _use_right_truncation_nowcasting_auto(
     *,
     target: str,
@@ -182,6 +183,7 @@ def _use_right_truncation_nowcasting_auto(
         True if right-truncation nowcasting should be used, False otherwise
     """
     return target == "nssp" and ed_visit_type in ["observed", "other"]
+
 
 def _use_right_truncation_nowcasting_right_truncation(
     *,
@@ -209,15 +211,14 @@ def _use_right_truncation_nowcasting_right_truncation(
     """
     if target != "nssp":
         raise ValueError(
-            "right-truncation nowcasting is currently supported only "
-            "for target='nssp'."
+            "right-truncation nowcasting is currently supported only for target='nssp'."
         )
     if ed_visit_type == "pct":
         raise ValueError(
-            "right-truncation nowcasting is not supported for "
-            "ed_visit_type='pct'."
+            "right-truncation nowcasting is not supported for ed_visit_type='pct'."
         )
     return True
+
 
 def _generate_right_truncation_nowcast(
     *,
@@ -229,8 +230,7 @@ def _generate_right_truncation_nowcast(
     param_data_dir: Path | str | None,
     logger: logging.Logger,
 ) -> NsspRightTruncationNowcast:
-    """Generate a right-truncation nowcast source for EpiAutoGP.
-    """ 
+    """Generate a right-truncation nowcast source for EpiAutoGP."""
     if frequency != "daily":
         logger.warning(
             "Using right-truncation nowcasting for frequency=%r. Confirm that "
@@ -254,6 +254,7 @@ def _generate_right_truncation_nowcast(
         )["right_truncation_pmf"]
 
     return NsspRightTruncationNowcast(right_truncation_pmf=right_truncation_pmf)
+
 
 def _resolve_nowcast_source(
     *,
--- a/pipelines/epiautogp/epiautogp_forecast_utils.py
+++ b/pipelines/epiautogp/epiautogp_forecast_utils.py
@@ -10,13 +10,13 @@ import os
 from dataclasses import dataclass
 from datetime import date
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Literal, get_args
 
 import polars as pl
 
 from pipelines.data.prep_data import get_pmfs, process_and_save_loc_data
 from pipelines.epiautogp.nowcast import NowcastSource
-from pipelines.epiautogp.nssp_nowcast import NsspRightTruncationNowcast
+from pipelines.epiautogp.reporting_delay_nowcast import ReportingDelayNowcast
 from pipelines.utils.common_utils import (
     append_prop_data_to_combined_data,
     calculate_training_dates,
@@ -28,7 +28,8 @@ from pipelines.utils.common_utils import (
     model_fit_dir_to_hub_tbl,
 )
 
-NowcastSourceName = Literal["auto", "none", "right-truncation"]
+NowcastSourceName = Literal["none", "reporting-delay"]
+VALID_NOWCAST_SOURCE_NAMES: tuple[str, ...] = get_args(NowcastSourceName)
 
 
 @dataclass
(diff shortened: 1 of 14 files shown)


pipelines/epiautogp/nssp_nowcast.py
```diff
@@ -0,0 +1,67 @@
...
+            for fraction in right_truncation_cdf
+            if fraction < 1.0 - REPORTING_FRACTION_TOL
+        ]
+
+        n_nowcast = min(len(reports), len(incomplete_cdf))
+        if n_nowcast == 0:
+            return NowcastData()
+
+        # Well behaved estimate when cases are zero, see:
+        # https://baselinenowcast.epinowcast.org/articles/model_definition.html#point-nowcast-generation
```
damonbayer: Nice! I would actually love to expose this somewhere more accessible so it could be used in https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1076

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -153,6 +158,156 @@ def post_process_forecast(self) -> None:
...
+            "right-truncation nowcasting is currently supported only for target='nssp'."
+        )
+    if ed_visit_type == "pct":
+        raise ValueError(
+            "right-truncation nowcasting is not supported for ed_visit_type='pct'."
+        )
+    return True
+
+
+def _generate_right_truncation_nowcast(
```
damonbayer: I think most of the instances called `right_truncation` could have a better name. Something that references `pmf` or `cdf`?

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -153,6 +158,156 @@ def post_process_forecast(self) -> None:
...
+    if nowcast_source_name not in valid_sources:
+        raise ValueError(
+            f"nowcast_source_name must be one of {valid_sources}, "
+            f"got {nowcast_source_name!r}"
+        )
+
+    if nowcast_source_name == "none":
+        return None
+
+    use_right_truncation = False
```
damonbayer: A little bit cleaner to move this to a final `else` block rather than initializing here and overwriting?

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -153,6 +158,156 @@ def post_process_forecast(self) -> None:
...
+    if nowcast_source_name == "auto":
+        use_right_truncation = _use_right_truncation_nowcasting_auto(
+            target=target,
+            ed_visit_type=ed_visit_type,
+        )
+    elif nowcast_source_name == "right-truncation":
+        use_right_truncation = _use_right_truncation_nowcasting_right_truncation(
+            target=target,
+            ed_visit_type=ed_visit_type,
+        )
```
damonbayer: Feels like this whole block should be `use_right_truncation(nowcast_source_name, target, ed_visit_type)` rather than the separate `_use_right_truncation_nowcasting_(target, ed_visit_type)` functions.

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -153,6 +158,156 @@ def post_process_forecast(self) -> None:
...
+    report_date: date,
+    param_data_dir: Path | str | None,
+    nowcast_source_name: NowcastSourceName,
+    right_truncation_pmf: list[float] | None,
+    logger: logging.Logger,
+) -> NowcastSource | None:
+    """
+    Resolve the requested nowcast source for one EpiAutoGP run.
+    """
+    valid_sources = ["auto", "none", "right-truncation"]
```
damonbayer: This and `NowcastSourceName = Literal["auto", "none", "right-truncation"]` should be linked up somehow so you don't have to change it in two places when you add a source.

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -153,6 +158,156 @@ def post_process_forecast(self) -> None:
...
+        Target data type (e.g., "nssp", "nhsn")
+    ed_visit_type : str
+        Type of ED visits (e.g., "observed", "other", "pct")
+
+    Returns
+    -------
+    bool
+        True if right-truncation nowcasting should be used, False otherwise
+    """
+    if target != "nssp":
```
damonbayer: I'm a bit lost at the moment. How does this relate to weekly vs daily nssp?

SamuelBrand1:
The core problem behind @damonbayer comments (I think!) is that the one nowcasting implementation in this PR doesn't apply to percentage data which has lead to some clunky barrier code. I'll have a rethink.

NB: Obviously applying a simplistic nowcast correction to denominator and numerator will either precisely cancel out (or approximately depending on precise implementation) so is pointless. So to be a useful contribution you'd need different right truncation pmfs for different aspects (other, target pathogens etc).

SamuelBrand1:
I've smoothed out the interface to nowcasting options. The main theme is to use the abstractions to reduce the crazy conditional branching that was v hard to parse and inelegant.

**Main changes**:

- The `NowcastSource` Protocol now requites an `applies_to` method. This should return `True` if the type of nowcasting problem we are interested in matches up with the nowcasting approach (e.g. don't nowcast %s with a singe right-truncation pmf).
- Renamed the basic right-truncation application nowcast to `ReportingDelayNowcast` because its more generic than "NSSP".
- Using `ReportingDelayNowcast` with epiweekly NSSP data felt like an edge case that was worth warning about, the most common way we are going to get a right-truncation pmf is from NNH which applies to daily. However, its not invalid to use the same idea on the weekly scale. I've moved the warning to being owned by the nowcasting method.
- Moved to using `match` to "dispatch" (its not really but you know) the string that comes in from command line to a nowcast implementation. The basic logic so far `none` -> no nowcasting and `reporting-delay` -> `ReportingDelayNowcast` `applies_to` (checks this is a valid nowcasting approach) and if that rets `True` does the simple nowcast.
- Removed the "auto" option... just lead to more conditional branching nonsense.

**minor change**

Moved the basic nowcast inflation funcs to `pipelines/epiautogp/reporting_delay.py` so importable by @damonbayer in other projects.

EDIT: by the way the simple inflation functions are also python lists of floats operations. Reasons: first, keeps typing simple (a nowcast is a list of dates and a list of lists of floats). Second, avoids a list -> jax array -> list round trip which would have been fine but read weirdly.

pipelines/epiautogp/nssp_nowcast.py
SamuelBrand1: Moved it into a new file pipelines/epiautogp/reporting_delay.py so you can import away!

pipelines/epiautogp/epiautogp_forecast_utils.py
SamuelBrand1: actually went for reporting delay... and variation on that. I feel that is clearer but open to suggestions.

pipelines/epiautogp/epiautogp_forecast_utils.py
SamuelBrand1: This branching was getting hellish... did a big smooth out.

pipelines/epiautogp/epiautogp_forecast_utils.py
SamuelBrand1: Good catch... I've aimed to make this source of truth so if its not thats a mistake

pipelines/epiautogp/epiautogp_forecast_utils.py
SamuelBrand1: Too many branches! I've smoothed this out

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -153,6 +159,78 @@ def post_process_forecast(self) -> None:
...
+            param_estimates=param_estimates,
+            loc_abb=loc,
+            disease=disease,
+            as_of=report_date,
+        )["right_truncation_pmf"]
+
+    return ReportingDelayNowcast(reporting_delay_pmf=reporting_delay_pmf)
+
+
+def _resolve_nowcast_source(
```
damonbayer, line 182: I think this can be a function of `target`, `ed_visit_type`, and `frequency` ? Since different sources will require different args, we should pass those through `kwargs`?

damonbayer: I think you could apply this idea throughout. We don't want the function signatures getting out of hand as we add more methods.

pipelines/epiautogp/nowcast.py
```diff
@@ -0,0 +1,73 @@
...
+    def get_nowcast_data(
+        self,
+        *,
+        dates: list[dt.date],
+        reports: list[float],
+    ) -> NowcastData:
+        """
+        Estimate nowcast data for one model run.
+        """
+        ...
```
damonbayer, line 51: I'm confused by these. It appears they take several arguments, but none of the arguments get used?

damonbayer: Oh I see they aren't used in `FixedNowcast`, but are used in `ReportingDelayNowcast`.

pipelines/epiautogp/epiautogp_forecast_utils.py
SamuelBrand1: Good catch, that was a leftover from the complicated branching first pass

SamuelBrand1: Actually, looking more carefully... going the kwarg route would erase the various constraints we've put in e.g. `Path | str | None` which opens up a different set of potential problems. However we do have groups of the same arguments getting passed around so this does suggest grouping into a class.

pipelines/epiautogp/nowcast.py
SamuelBrand1: added a docstring

pipelines/epiautogp/epiautogp_forecast_utils.py
SamuelBrand1: So what I've done is bundled the values that _specify_ the forecast into ForecastSpec dataclass. And kept the  nowcast_source_name. But the arguments _to_ the nowcast constructor splat through.

SamuelBrand1:
@damonbayer next steps to smooth this approach:

- Part of the large number of args getting passed around was that things like "target" "loc" etc get passed around as groups. I rationalised this by adding a `ForecastSpec` dataclass which keeps them together.
- Updated everywhere they get groups to use this (except the `ForecastPipelineContext` constructor where all the args pass in).
- For the nowcasting constructor the forecast spec and the name of the nowcasting approach get type hints, all the other args (which will differ from nowcast type to nowcast type) get splatted at kwargs.

I think this simplifies things whilst leaving lots of room to extend this interface e.g. `nowcastnhsn` and any other future nowcasting approach we want.

damonbayer (approved):

PR #1084, opened 2026-05-20
Require a manual `workflow_dispatch` when you want to push to dagster. 

jkislin:
- Pushes and PRs will never trigger a push to the dagster server.
- Workflow dispatch is the only thing that can trigger a push to the dagster server, and you have to specify that's what you want.


jkislin:
Need to rebase.

PR #1087, opened 2026-05-22
Issue 1058: nhsn nowcast ingestion

### LINKED ISSUE #1058: NHSN nowcast ingestion

SamuelBrand1:
This pull request introduces a new `NowcastSource` for Hubverse formatted nowcasting samples.

The idea is to pass a URI to a handoff pointer json, this contains all the information about a given nowcasting prod run so we can validate that we are connecting to the correct prod run. The pointer also contains further URIs about where to find the actual hubverse formatted nowcasts.

~**This is WIP until running live tests, but here for @damonbayer @dylanhmorris review**~

**Hubverse Nowcast Integration**

- Added `HubversePointerNowcast` as a pluggable nowcast source, allowing the pipeline to read negative-horizon nowcast samples from a Hubverse handoff pointer JSON whose `hubverse.model_output_uri` points to a sample-format parquet file.
- Updated pipeline utility functions and the main forecast entry point to support the new `hubverse` nowcast source and to accept the `--hubverse-nowcast-pointer-uri` argument (`pipelines/epiautogp/epiautogp_forecast_utils.py`, `pipelines/epiautogp/forecast_epiautogp.py`) 

**Documentation and Usage**

- Documentation to describe the new Hubverse nowcast source, including its configuration, pointer requirements, and usage examples in both the main and utility README files.
- Added instructions and a helper script for mounting only the containers required by a Hubverse handoff pointer, supporting local, file, and Azure Blob (`az://`) URIs c.f. `blobfuse/README.md`, `blobfuse/mount_hubverse_pointer.sh`.

**Test Data Generation**

- Added a utility function and test data generation logic to produce local Hubverse-format nowcast artifacts and handoff pointers for testing and development (`pipelines/data/generate_test_data.py`, `pipelines/data/generate_test_data_lib.py`).

These changes make it possible to use production Hubverse nowcasts as direct inputs to the EpiAutoGP pipeline, improving interoperability and reproducibility across nowcasting workflows.


blobfuse/mount_hubverse_pointer.sh
damonbayer, line 1: I'm pretty opposed to any additional functionality in this blobfuse directory. The directory is really only here because @jkislin sometimes uses it for local dagster testing. Can we just add the relevant blob storage container to the list of mounts and work with that?

SamuelBrand1:
Swapped to mounting:

- Deleted blobfuse/mount_hubverse_pointer.sh and the matching blobfuse/README.md section.
- Added `nowcastnhsn-prod` (intended blob container rather than nowcastnhsn-experiment) to the static TO_MOUNT list in blobfuse/mount.sh (with the usual symlink stuff).
- Added the corresponding volume bind to both the Docker executor and Azure Batch executor lists in dagster_defs.py.
- Cleaned up the pipelines/epiautogp/README.md section to drop the helper-script instructions

SamuelBrand1:
I've backed off any changes to `blobfuse`.

I've also updated the testing so `pipelines/tests/test_epiautogp_nhsn_nowcast_vm.py` now only works from VAP, but it will run using the standard data path for NHSN data and the existing nowcasting in `nowcastnhsn-prod` blob. I've tried to model this test function on the run functions in dagster-defs.py.

pipelines/tests/test_epiautogp_nhsn_nowcast_vm.py
```diff
@@ -0,0 +1,218 @@
...
+)
+from pipelines.epiautogp.forecast_epiautogp import run_epiautogp_forecast
+from pipelines.epiautogp.hubverse_pointer import resolve_artifact_path
+
+RUN_ENV = "EPIAUTOGP_NHSN_NOWCAST_VM_TEST"
+TARGET_ENV = "EPIAUTOGP_NHSN_NOWCAST_VM_TARGET"
+LOCATION_ENV = "EPIAUTOGP_NHSN_NOWCAST_VM_LOCATION"
+CONTAINER_ENV = "EPIAUTOGP_NHSN_NOWCAST_VM_CONTAINER"
+POINTER_URI_ENV = "EPIAUTOGP_NHSN_NOWCAST_VM_POINTER_URI"
+FACILITY_NSSP_DIR_ENV = "EPIAUTOGP_NHSN_NOWCAST_VM_NSSP_DATA_DIR"
```
damonbayer: I don't find these names descriptive enough to know what these are supposed to be.

Add them to [`.env.example`](https://github.com/CDCgov/cfa-stf-routine-forecasting/blob/main/.env.example)?

pipelines/tests/test_epiautogp_nhsn_nowcast_vm.py
```diff
@@ -0,0 +1,218 @@
...
+    "rsv": "RSV",
+}
+
+
+def _skip_unless_enabled() -> None:
+    if os.environ.get(RUN_ENV) != "1":
+        pytest.skip(f"Set {RUN_ENV}=1 on a VM with nowcastnhsn-prod mounted to run.")
+
+
+def _require_r_packages(*packages: str) -> None:
```
damonbayer: I don't think this is needed. Especially not in such a specific location. If you are missing R or a required package, I expect the OS or R to provide a readable error without needing a wrapper.

pipelines/epiautogp/README.md
```diff
@@ -54,15 +56,37 @@ Shared utilities for the forecast pipeline, containing modular functions for eac
...
+  --nowcast-source hubverse \
+  --hubverse-nowcast-pointer-uri az://nowcastnhsn-prod/nowcastnhsn-prod/latest/covid.json \
+  ...
+```
+
+The pointer must be a production `handoff_pointer` JSON whose `target` and `report_date` match the EpiAutoGP run and whose `hubverse.validation_status` is `passed`. The source then reads `hubverse.model_output_uri`.
+
+The Hubverse row target is inferred from `ForecastSpec`: disease maps to `covid`/`flu`/`rsv`, `target` and `ed_visit_type` choose the producer variable, and `frequency` adds the epiweekly `wk ` prefix. The inference mirrors the producer convention in `hewr/R/to_hubverse_tbl.R`.
+
+Supported URI forms are local paths, `file://` URIs, and `az://container/blob` URIs. `az://` paths are resolved through this repo's existing blobfuse mounts (`./blobfuse/mounts/<container>/...` or `/mnt/<container>/...`). The `nowcastnhsn-prod` container is included in `blobfuse/mount.sh`'s static mount list, so `make mount` makes the pointer reachable. For tests, provide a local or `file://` URI.
```
damonbayer: > The `nowcastnhsn-prod` container is included in `blobfuse/mount.sh`'s static mount list

I don't think it is?

pipelines/epiautogp/hubverse_pointer.py
```diff
@@ -0,0 +1,211 @@
...
+    candidates: list[Path] = []
+    for root in roots:
+        candidate = root / container / blob_path
+        if candidate not in candidates:
+            candidates.append(candidate)
+    return candidates
+
+
+def _blobfuse_path(raw_uri: str, *, source_label: str) -> Path:
+    """Resolve an az:// URI to an existing blobfuse-mounted path."""
```
damonbayer: Why do we need to support parsing `az://`? The point of mounting the storage container is that we can just work with the file path.

pipelines/epiautogp/hubverse_pointer.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+
+
+def resolve_artifact_path(
+    raw_uri: str | Path,
+    *,
+    base_path: Path | None = None,
+    source_label: str,
+) -> Path:
+    """Resolve local, file://, and blobfuse-backed az:// URIs to a local path."""
```
damonbayer: Why do we need to support parsing `file://`? The point of mounting the storage container is that we can just work with the file path.

pipelines/data/generate_test_data_lib.py
```diff
@@ -90,6 +90,89 @@
...
+        )
+        .sort("weekendingdate")
+        .tail(3)
+    )
+    if source.height < 2:
+        raise ValueError(
+            f"Need at least 2 NHSN rows for {disease} {location} Hubverse nowcasts."
+        )
+
+    rows = [
```
damonbayer, line 133: Put `pl.DataFrame` around this and give it a meaningful name.

pipelines/epiautogp/hubverse_nowcast.py
```diff
@@ -0,0 +1,304 @@
...
+
+    @classmethod
+    def from_pointer_uri(
+        cls,
+        *,
+        pointer_uri: str | Path | None,
+        forecast_spec: ForecastSpec,
+    ) -> HubversePointerNowcast:
+        """Build a Hubverse nowcast source from a required pointer URI."""
+        if pointer_uri is None:
```
damonbayer: in the type hints you say `pointer_uri` can be `None`, but here you error if it is `None`?

pipelines/data/generate_test_data.py
```diff
@@ -336,6 +337,16 @@ def main(base_dir: Path, clean: bool):
...
 
+    hubverse_nowcast_dir = Path(private_data_dir, "hubverse_nowcasts")
+    for name, data in nhsn_data_combined.group_by("disease", "jurisdiction"):
+        write_hubverse_nhsn_nowcast_artifacts(
+            hubverse_nowcast_dir,
+            nhsn_data=data,
+            disease=name[0],
+            location=name[1],
+            report_date=max_train_date,
+        )
```
damonbayer:
```suggestion
    for (disease, location), data in nhsn_data_combined.group_by("disease", "jurisdiction"):
        write_hubverse_nhsn_nowcast_artifacts(
            hubverse_nowcast_dir,
            nhsn_data=data,
            disease=disease,
            location=location,
            report_date=max_train_date,
        )
```

damonbayer (changes requested):
A few comments after initial review. The scaffolding around epiautogp is quite complex and I don't yet understand it all.

pipelines/epiautogp/README.md
SamuelBrand1: Thats left over from backing out mount.sh change

pipelines/epiautogp/hubverse_pointer.py
SamuelBrand1: The `az://` scheme provided mount-discovery by searching `./blobfuse/mounts/`, `/mnt/`, repo root, etc. for the mounted container, and `file://` did URI semantics. This was just a defensive thing: callers didn't need to know the exact mount path. But I'll get rid of it to simplify

SamuelBrand1:
Next batch of smoothing this out:

- Removed az:// and file:// URI resolution. 
- Fixed from_pointer_uri() type hint from str | Path | None to str | Path, removed the pointless None guard.
- Moved the None validation to _build_hubverse_nowcast() where it belongs
- Simplified CLI help text getting rid of the az:// and file:// comments
- Replaced the 6 custom env vars to choose the end-to-end vap only test and _skip_unless_enabled() and replaced with @requires_ext_catalog. Removed _require_r_packages() as a overkill check
- Removed 3 tests for file:// and az:// URI resolution (no longer supported).
- Named the anonymous DataFrame as nowcast_samples.
- Applied (disease, location), data destructuring and merged the two group_by loops into one.
- In the epiautogp README removed mistaken ref to mount.sh.

SamuelBrand1:
> A few comments after initial review. The scaffolding around epiautogp is quite complex and I don't yet understand it all.

epiautogp needs to get some nowcasts so I give it a pointer json file.

The pointer json is mainly for validation gate and safety checking. The pointer's hubverse.validation_status field is checked to be "passed" before the parquet is read, this is the output of doing `hubValidation` at the end of the nowcastNHSN production to check that the outputs make sense. And the pointer's report_date is validated against the forecast spec. 

Then it tells epiautogp where to find the appropriate nowcasts which go in, and get used by the model to do forecasting.

pipelines/epiautogp/README.md
damonbayer: I thought the resolution to my comment would be adding `nowcastnhsn-prod` to `mount.sh`? That would be the simplest way to getting your VAP-only test to work, right?

SamuelBrand1: Right, when you said you were "pretty opposed to any additional functionality in this blobfuse directory," I read that as "don't touch blobfuse at all" rather than "don't add a separate script — use the existing mount pattern." So I backed off entirely.

What I've done now: added nowcastnhsn-prod to the TO_MOUNT list in mount.sh with the standard symlink block. The VM only end-to-end test now reads the pointer through that symlink path rather than using the custom env vars.

SamuelBrand1:
@damonbayer I got mixed up about your request to not add functionality as "don't change anything in /blobfuse"... put the prod container for nowcasting in

pipelines/epiautogp/README.md
damonbayer: That's exactly what I envisioned! Sorry for the confusion.

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -179,6 +180,23 @@ def _build_reporting_delay_nowcast(
     return ReportingDelayNowcast(reporting_delay_pmf=reporting_delay_pmf)
 
 
+def _build_hubverse_nowcast(
+    *,
+    forecast_spec: ForecastSpec,
+    hubverse_nowcast_pointer_uri: Path | str | None,
+) -> HubversePointerNowcast:
+    """Build a generic Hubverse handoff-pointer nowcast source."""
+    if hubverse_nowcast_pointer_uri is None:
```
damonbayer: Why indicate `None` as an acceptable type if it leads to an error?

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -179,6 +180,23 @@ def _build_reporting_delay_nowcast(
     return ReportingDelayNowcast(reporting_delay_pmf=reporting_delay_pmf)
 
 
+def _build_hubverse_nowcast(
```
damonbayer: I don't think we need this function. It just takes two arguments and simply passes them to `HubversePointerNowcast.from_pointer_uri`. Just use `HubversePointerNowcast.from_pointer_uri` instead?

pipelines/epiautogp/forecast_epiautogp.py
```diff
@@ -166,9 +167,12 @@ def main(
     n_threads : int | str, default="auto"
         Number of threads for Julia execution (integer or "auto")
     nowcast_source_name : str, default="none"
-        Nowcast source to use: "none" or "reporting-delay"
+        Nowcast source to use: "none", "reporting-delay", or "hubverse"
     reporting_delay_pmf : list[float] | None, default=None
         Directly supplied reporting-delay PMF. Python API only.
```
damonbayer, line 175: What does `Python API only` indicate?

pipelines/epiautogp/forecast_epiautogp.py
```diff
@@ -106,6 +106,7 @@ def main(
     param_data_dir: Path | str = Path("private_data", "prod_param_estimates"),
     nowcast_source_name: str = "none",
```
damonbayer: `Literal` type hint to make the options explicit?

pipelines/epiautogp/forecast_epiautogp.py
```diff
@@ -106,6 +106,7 @@ def main(
...
     reporting_delay_pmf: list[float] | None = None,
+    hubverse_nowcast_pointer_uri: Path | str | None = None,
```
damonbayer: Having  separate args for `nowcast_source_name`, `reporting_delay_pmf`, and `hubverse_nowcast_pointer_uri` feels bad, when you know that you need to match up certain source names with these other args. Seems ripe for a `nowcast_config` object or something. Perhaps over-engineered for right now.

pipelines/epiautogp/forecast_epiautogp.py
```diff
@@ -166,9 +167,12 @@ def main(
     n_threads : int | str, default="auto"
         Number of threads for Julia execution (integer or "auto")
     nowcast_source_name : str, default="none"
-        Nowcast source to use: "none" or "reporting-delay"
+        Nowcast source to use: "none", "reporting-delay", or "hubverse"
     reporting_delay_pmf : list[float] | None, default=None
         Directly supplied reporting-delay PMF. Python API only.
+    hubverse_nowcast_pointer_uri : Path | str | None, default=None
```
damonbayer: The `pointer_uri` naming is a bit confusing. This is meant to be a file path, right?

pipelines/epiautogp/hubverse_pointer.py
```diff
@@ -0,0 +1,153 @@
...
+from typing import Any
+
+from pipelines.epiautogp.forecast_spec import ForecastSpec
+
+
+def resolve_artifact_path(
+    raw_path: str | Path,
+    *,
+    base_path: Path | None = None,
+    source_label: str,
```
damonbayer: `source_label` is unused?

pipelines/epiautogp/hubverse_pointer.py
```diff
@@ -0,0 +1,153 @@
...
+) -> Path:
+    """Resolve a local file path (absolute or relative) to a concrete path."""
+    path = Path(raw_path)
+    if path.is_absolute():
+        return path.resolve()
+    if base_path is not None:
+        return (base_path.parent / path).resolve()
```
damonbayer: I'm a bit confused about the arguments here. Why do we pass `base_path` and expect to use its parent? Why not just expect the parent path to be passed?

pipelines/epiautogp/forecast_epiautogp.py
SamuelBrand1: You can debug this by running from python runtime with a custom pmf

pipelines/epiautogp/hubverse_pointer.py
```diff
@@ -0,0 +1,153 @@
...
+    return (Path.cwd() / path).resolve()
```
damonbayer: Why not use `path.absolute()`?

pipelines/epiautogp/forecast_epiautogp.py
SamuelBrand1: But its not a CL option (although you could save a file to point at)

pipelines/epiautogp/hubverse_pointer.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def _load_pointer(
+    pointer_uri: str | Path,
+    *,
+    source_label: str,
+) -> tuple[dict[str, Any], Path]:
+    """Read a handoff pointer JSON and return it with its resolved path."""
+    resolved = resolve_artifact_path(pointer_uri, source_label=source_label)
```
damonbayer: source_label isn't used in `resolve_artifact_path`, therefore it isn't used here?

pipelines/epiautogp/hubverse_pointer.py
damonbayer, line 1: I think this file could use a once over @SamuelBrand1. I'm not sure what's going on with the `source_label`.

pipelines/epiautogp/hubverse_nowcast.py
```diff
@@ -0,0 +1,299 @@
...
+
+def _target_prefix(frequency: str) -> str:
+    if frequency == "epiweekly":
+        return "wk "
+    if frequency == "daily":
+        return ""
+    raise ValueError(
+        f"Unsupported Hubverse nowcast frequency {frequency!r}; "
+        "expected 'daily' or 'epiweekly'."
+    )
```
damonbayer, line 65: Not suggesting to do this in this PR, but I think it might be an improvement to shift epiautogp to work with "target" names directly, rather than parsing the various "frequency", "disease", and "ed_visit_type" args.

damonbayer (changes requested):

pipelines/epiautogp/forecast_epiautogp.py
SamuelBrand1: I've left it like this way, but there is room to add other nowcast approaches by extending the Protocol... thats when I'd revisit this (if we ever do that)

SamuelBrand1:
Good spots. The "URI" naming was leftover from when `resolve_artifact_path` parsed `az://` and `file://` schemes. Now that it only takes plain paths.

- **`resolve_artifact_path` got rid of unused parameter and replaced `(Path.cwd() / path).resolve()` with `path.absolute()`.
- `pointer_uri` → `pointer_path` everywhere.
- Dropped `_build_hubverse_nowcast`. Inlined the `None` check and `HubversePointerNowcast(...)` construction into the `"hubverse"` case of `_resolve_nowcast_source`. Also dropped the now-orphaned `HubversePointerNowcast.from_pointer_path` factory since the inline uses the constructor directly.
- **`Literal` on `nowcast_source_name`.**

pipelines/epiautogp/hubverse_pointer.py
SamuelBrand1: `source_label` was a capture for error messages but then I forgot about it!

pipelines/epiautogp/hubverse_nowcast.py
SamuelBrand1: Isn't it better to have one source of truth about this run? E.g. the specification of the forecast run forces the target col elements so if its wrong thats good?

SamuelBrand1:
Refactored `hubverse_pointer.py` to collapse all the helpers into one function and I think its more obvious what its doing now. I put in this _expects checker/helper but maybe I should publish a JSON Schema for the handoff pointer? Then we can do a direct checker

SamuelBrand1:
Dropped `source_label` from `hubverse_nowcast.py` too now. The default source label "Hubverse nowcast" was never actually overridden. I wanted to capture more things in case of errors but I think it just got unwieldly.

pipelines/epiautogp/epiautogp_forecast_utils.py
```diff
@@ -231,6 +248,7 @@ def setup_forecast_pipeline(
     param_data_dir: Path | str | None = None,
     nowcast_source_name: NowcastSourceName = "none",
```
damonbayer: Why `"none"` instead of `None`?

SamuelBrand1: I liked the intentionality of three options with not doing nowcast being a choice... but its cleaner pushing None through.

SamuelBrand1: Swapped this to None now

damonbayer:
@SamuelBrand1 After our discussion today, I arrived at https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1110 as a good first step for this. We just need a nice interface for accessing the nowcastNHSN results. We already have examples of nice interfaces for accessing data in `cfa.stf.data`, so we should build on to that part of the code base, I think.

PR #1092, opened 2026-05-26
cast `geo_value` in `get_nssp` as string

### LINKED ISSUE #1090: Schema mismatch when loading `nssp_comprehensive` with `cfa.stf.data.get_nssp()`
```
get_nssp(
    disease=["COVID-19", "Total"],
    loc_abb="US",
    dataset="comprehensive",
)
```
Error snippet:
```
polars.exceptions.InvalidOperationError: 'union'/'concat' inputs should all have the same schema,got
Schema { fields: {"report_date": Date, "reference_date": Date, "metric": String, "geo_type": String, "geo_value": String, "disease": String, "value": Float64}, metadata: () } and 
Schema { fields: {"report_date": Date, "reference_date": Date, "metric": String, "geo_type": String, "geo_value": Categorical, "disease": String, "value": Float64}, metadata: () }
```
Cause of the error: `get_nssp` casts `geo_value` to a categorical but `geo_value` column in `comprehensive_nssp_gold` are strings 

cc @O957 for awareness

sbidari:


cfa/stf/data/get_data.py
```diff
@@ -272,11 +272,13 @@ def get_nssp(
...
     )
 
+    geo_value_dtype = dat.schema["geo_value"]
+
     combined_dat = (
         pl.concat(
             [
                 state_dat,
-                dat.with_columns(pl.lit("US").alias("geo_value").cast(pl.Categorical)),
+                dat.with_columns(pl.lit("US").alias("geo_value").cast(geo_value_dtype)),
```
damonbayer: It would be better to have a consistent output format (always categorical or always string, regardless of the value of `dataset`). Slightly prefer string, I think.

damonbayer (changes requested):

cfa/stf/data/get_data.py
dylanhmorris: Agree, I think we want string here.

sbidari: done https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1092/commits/716f415a55973cabe5f98a869031c668b18aa08b

damonbayer (changes requested):
@sbidari please add a test to `tests/cfa/stf/data/test_get_nssp.py` related to this. At a minimum, add something to the `@requires_ext_catalog` section that tests that we can specify the comprehensive dataset. You're welcome to be more thorough, if you want.

damonbayer (approved):

PR #1117, opened 2026-06-16
Set Azure HTTP Policy Logs to WARNING level; reduce clutter in model output

jkislin:
By setting the Azure Core HTTP Policy Logger to `WARNING`, we can suppress unnecessarily verbose logs in our model outputs.

### Before, with Azure HTTP Logs:
image

### After, with only app logs:
image


damonbayer (approved):
Thanks @jkislin

PR #1154, opened 2026-06-30
cleanup dead R code

damonbayer:


pipelines/utils/create_prop_samples.R
dylanhmorris, line 8: Was/is this function used anywhere? Based on the diff, it appears it was previously (incorrectly?) returning nothing.

If this was not an issue (b/c it was not in use), should we just remove?

dylanhmorris (approved):
LGTM. One question.

pipelines/utils/create_prop_samples.R
damonbayer: The syntax used here does an invisible return, so there is no change to the functionality

``` r
my_func <- function(a) {
  b <- a +1
}
val <- my_func(0)
val
#> [1] 1
```

<sup>Created on 2026-06-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

dylanhmorris: TIL a final line assignment returns the value.

PR #1165, opened 2026-07-09
Use Python append prop data helper

damonbayer:
## Summary

Replaces the `append_prop_data.R` call path with Python file I/O in `append_prop_data_to_combined_data()`, using the new `cfa.stf.forecasttools.append_prop_data()` helper.

The routine pipeline keeps ownership of reading and writing combined-data files, formats CSV/TSV `.value` output to match the legacy `readr::write_tsv` representation, and leaves parquet output numeric.

This PR depends on CDCgov/cfa-stf-forecasttools#221 being merged first.

## Validation

- `uv run python -c "import sys; sys.path.insert(0, '/Users/damon/Documents/GitHub/forecasttools-py'); import pytest; raise SystemExit(pytest.main(['pipelines/tests/utils/test_common_utils.py']))"`
- Compared legacy R script output with the new Python routine function on the same TSV fixture; outputs were byte-identical and parsed tables matched.

### DIFF
--- a/pipelines/tests/utils/test_common_utils.py
+++ b/pipelines/tests/utils/test_common_utils.py
@@ -5,8 +5,8 @@ import datetime as dt
 import logging
 
 import polars as pl
-import polars.testing as plt
 import pytest
+from polars.testing import assert_frame_equal
 
 from pipelines.utils.cli_utils import (
     add_common_forecast_arguments,
@@ -162,7 +162,7 @@ class TestDataWranglingUtils:
                 ".value": [2.0, 8.0, 0.2],
             }
         )
-        plt.assert_frame_equal(result, expected)
+        assert_frame_equal(result, expected)
 
     def test_append_prop_data_to_combined_data_allows_variable_names(self, tmp_path):
         data_path = tmp_path / "combined_data.tsv"
@@ -195,7 +195,7 @@ class TestDataWranglingUtils:
                 ".value": [7.0, 3.0, 0.3],
             }
         )
-        plt.assert_frame_equal(result, expected)
+        assert_frame_equal(result, expected)
 
 
 class TestCLIUtils:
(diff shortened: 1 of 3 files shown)


pipelines/tests/utils/test_common_utils.py
```diff
@@ -4,13 +4,16 @@
 import datetime as dt
 import logging
 
+import polars as pl
+import polars.testing as plt
```
dylanhmorris: Suggest this since only one function is used and `plt` is the standard shortening for `matplotlib.pyplot`
```suggestion
from polars.testing import assert_frame_equal
```

pipelines/tests/utils/test_common_utils.py
```diff
@@ -133,6 +136,67 @@ def test_get_available_reports_with_parquet_files(self, tmp_path):
...
+                "location": ["US", "US", "US"],
+                ".variable": [
+                    "observed_ed_visits",
+                    "other_ed_visits",
+                    "prop_disease_ed_visits",
+                ],
+                ".value": [2.0, 8.0, 0.2],
+            }
+        )
+        plt.assert_frame_equal(result, expected)
```
dylanhmorris:
```suggestion
        assert_frame_equal(result, expected)
```

pipelines/tests/utils/test_common_utils.py
```diff
@@ -133,6 +136,67 @@ def test_get_available_reports_with_parquet_files(self, tmp_path):
...
+                "location": ["US", "US", "US"],
+                ".variable": [
+                    "denom_other_visits",
+                    "num_visits",
+                    "prop_num_visits",
+                ],
+                ".value": [7.0, 3.0, 0.3],
+            }
+        )
+        plt.assert_frame_equal(result, expected)
```
dylanhmorris:
```suggestion
        assert_frame_equal(result, expected)
```

dylanhmorris (commented):
LGTM!

dylanhmorris (approved):

PR #1175, opened 2026-07-16
Add Support For Lookback Days Per Model

### LINKED ISSUE #1163: Per-model lookbacks
For COVID HE, we want to use 60 day lookback by default. For all other models we want to use 150 days by default. This requires us to run both the  60 and 150 days lookback for the Fable E models by default.

To close this issue, demonstrate:

1. Running the default set of models results in PyRenew-HE COVID (and all downstream fusion models) w/ 90 day lookback and all other models with 150 day lookback.
2. A rereun with unspecified a lookback results in the same as (1)
3. A rerun with specified lookback overrides the revised defaults.

O957:
This PR enables us to specify the lookback days of COVID-19 HE to be 60 rather than 150 days.


dagster_defs.py
```diff
@@ -277,11 +277,15 @@ def resolve_config(
         for entry in self.config_overrides:
             entry_dict = entry if isinstance(entry, dict) else entry.as_dict()
             if all(
-                key not in entry_dict or entry_dict[key] == value
+                entry_dict.get(key) is None or entry_dict.get(key) == value
```
dylanhmorris, line 282: Reason for not using
```python
                key not in entry_dict or entry_dict.get(key) == value
```
as you originally had?

Pretty sure these lines are equivalent due to short-circuit evaluation, and I think the intent of the former is a bit clearer.

dylanhmorris: With `.get` and defaults, one could even do this, but I think it hides (what I take to be) the intention even more

```suggestion
                entry_dict.get(key, value)  == value
```

jkislin:
A high level note - this adds somewhat to the complexity of `dagster_defs.py`, so I'm wondering - can any of this be added into the forecast modules themselves? If not that's alright, worth an ask though.

damonbayer (commented):
`dagster_defs` is getting quite complex. I wonder if this would be easier to understand if we broke it up. Not necessarily suggesting that for this PR.

dagster_defs.py
```diff
@@ -220,24 +222,73 @@ class ModelBaseConfig(_ModelTrainingFields, dg.ConfigurableResource):
...
-            ""
+            "Provide overrides as a list of dictionaries keyed by any combination of "
+            "location, disease, and model; an omitted key matches all values. "
+            "The Launchpad accepts both yaml and json-style lists e.g. "
+            "config_overrides: [{ disease: COVID-19, model: pyrenew_he, n_training_days: 60 }]"
         ),
     )  # type: ignore[reportInvalidTypeForm]
 
-    def get_by_location(self, loc: Location) -> "ModelBaseConfig":  # type: ignore[reportInvalidTypeForm]
+    def resolve_config(
```
damonbayer, line 240: I find this function pretty opaque. Maybe we could add a test to illustrate its behavior?

dagster_defs.py
```diff
@@ -220,24 +222,73 @@ class ModelBaseConfig(_ModelTrainingFields, dg.ConfigurableResource):
     # Add defaults here, or add in the launchpad with ctrl+space
```
damonbayer, line 222: Having a "override" that is on by "default" is fishy to me. I don't think understanding what happens by default should necessitate looking at an "override," but maybe this is the best solution available.

dylanhmorris: I think what we want here is a three-level hierarchy:

- Global defaults spanning models, optionally overridden every run by (e.g. "150 day lookback across models and locations")
- Specific defaults for subsets of model runs (e.g. "shorter lookback for COVID-19 HE models across locations"). These replace the global defaults.
- Manual (re)-configuration when the user triggers a run (e.g. "drop an extra day for GA COVID and MA RSV across models"). These replace the global defaults and any relevant specific defaults.

Agnostic on the best way to name / achieve this in code

dylanhmorris: Looking at the code, I think we're in a middle ground between two options, and we should choose one.

Right now, a base config is instantiated and then a `loc_config` (really a specific model run config) is instantiated from it within the forecast_x functions via `resolve_config()`.

https://github.com/CDCgov/cfa-stf-routine-forecasting/blob/898b9449a0e07de74f2dd1c3730d3a3f7fb57d86/dagster_defs.py#L442-L444

The actual call that runs the model then uses a _mix_ of `loc_config` and `model_base_config` values:

https://github.com/CDCgov/cfa-stf-routine-forecasting/blob/898b9449a0e07de74f2dd1c3730d3a3f7fb57d86/dagster_defs.py#L454-L462

I think better either to create the run config from the base config the top of the function call and then use it for _all_ parameters, or have methods for extracting parameters from the base config that are run-specific.

So in the snippet above we'd either have things like

```python
        n_training_days=run_config.n_training_days,
        fail_on_stale_data=run_config.fail_on_stale_data,
```
or

```python
        n_training_days=model_base_config.get("n_training_days", disease=disease, model=model, location=location)
        fail_on_stale_data=model_base_config.fail_on_stale_data,
``` 

or better yet if there's some way to configure `current_value` so that the run-specific values (if any) are determined upstream, we could just do

```python
        n_training_days=model_base_config.n_training_days.current_value,
        fail_on_stale_data=model_base_config.fail_on_stale_data.current_value,
```

I think my preference is for the last option if it's possible, otherwise the first (avoiding the middle), but defer to @giomrella @jkislin @damonbayer on what's possible and what's most dagster idiomatic.

giomrella: You should be able to do something like:
```
run_config = model_base_config.resolve_config( 
     disease=disease, location=location, model=f"pyrenew_{model_letters}" 
 ) 
```
And then just use `run_config` for all config values.

dylanhmorris: ^ yeah, that is what I was envisioning as option 1.

dylanhmorris: Let's do that (option 1) for now @O957 unless @damonbayer objects.

damonbayer:
Closing this since @jkislin is taking over.

PR #1212, opened 2026-08-05
CI experiment: verify urca dependency enforcement

damonbayer:
Temporary negative-control PR. Do not merge.\n\nThis removes urca from stfroutineforecasting/DESCRIPTION while pipelines/fable/fit_fable.R still explicitly loads it. The purpose is to verify that the production-like hard-dependency installation in GitHub Actions no longer obtains urca transitively and that CI catches the missing declaration. This PR will be closed after the result is captured.

### DIFF
--- a/stfroutineforecasting/DESCRIPTION
+++ b/stfroutineforecasting/DESCRIPTION
@@ -30,8 +30,7 @@ Imports:
     tibble,
     tidybayes,
     tidyr,
-    tidyselect,
-    urca
+    tidyselect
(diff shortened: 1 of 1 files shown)


damonbayer:
Negative control confirmed after deleting the accessible Ubuntu R caches: the clean-cache Fable rerun failed because urca was not installed. Evidence: https://github.com/CDCgov/cfa-stf-routine-forecasting/actions/runs/31033588424/job/92416594086

PR #1231, opened 2026-08-17
Unify forecast run state and pipeline orchestration

### LINKED ISSUE #1203: Consolidate EpiAutoGP specifics into unified ForecastData/Spec
I think the pythonic way would be to have a class which coordinates the behaviour of the pipeline, which is essentially set up for `pyrenew-hew` but each new model (if it wants to do different pipeline processing) needs to have subclass to extend/modify its methods.

_Originally posted by @SamuelBrand1 in https://github.com/CDCgov/pyrenew-hew/pull/783#discussion_r2620703726_

damonbayer:
## Summary

This PR gives the Fable, PyRenew, and EpiAutoGP single-location forecasts one shared pipeline lifecycle. It centralizes run state and common artifact handling while leaving model-specific preparation and execution in small, explicit hooks.

The refactor preserves the existing operational entry points, model names, source rules, forecast horizons, serialized JSON contracts, and output layout.

## What changed

- Added a frozen `ForecastRun` as the source of truth for one model, location, and forecast vintage, including dates and derived output paths.
- Added `SurveillanceInputs` to keep loaded surveillance frames, freshness metadata, and location population together.
- Added `ForecastPipeline.execute()` and migrated Fable, PyRenew, and EpiAutoGP to the shared lifecycle.
- Standardized the boundary between phases: preparation leaves executable inputs, model execution leaves `samples.parquet`, and publishing consumes those samples.
- Centralized common serialization, proportion augmentation, figure generation, Hubverse output creation, and lifecycle logging.
- Added contributor documentation for implementing and integrating another forecasting model.
- Consolidated forecast test fixtures and strengthened coverage of entry-point option forwarding, lifecycle ordering, derived run state, and model-specific input resolution.

## Pipeline lifecycle

After validating model configuration, every pipeline now follows the same four stages:

1. **Build the run** — calculate the training window, load the requested surveillance sources, and construct the canonical `ForecastRun`.
2. **Prepare inputs** — serialize common surveillance data, apply model-specific transformations, add derived proportions, and create any model-specific artifacts.
3. **Run the model** — execute the model and convert native output when needed so the phase always ends with a standardized `samples.parquet`.
4. **Publish outputs** — generate figures, credible intervals, and the model-level Hubverse table.

## Model-specific updates

- **EpiAutoGP:** moves optional epiweekly aggregation, nowcast resolution, JSON creation, and configuration validation into the shared preparation flow.
- **PyRenew:** resolves and serializes generation-interval, infection-to-admission, and right-truncation PMFs during preparation; reuses the run's population; and converts native output before `run_model()` returns. Prior setup now copies `priors.py` without maintaining redundant TOML metadata, allowing removal of `tomli-w`.
- **Fable:** adopts the same lifecycle and standardized sample/output contracts with minimal model-specific orchestration.

## Operational improvements

- Model entry points accept an injected logger so Dagster keeps its structured logging context.
- R and Julia subprocess output can stream during production runs while shared command helpers continue to capture output by default.
- Resolved resources are passed explicitly instead of being retained as mutable pipeline state.
- `right_truncation_offset` is derived from the dates on `ForecastRun` rather than stored with surveillance data.

## Validation

CI for the current PR head covers:

- Python tests and coverage checks
- Dagster definition validation
- Reduced pipeline end-to-end test
- Fable, PyRenew H/E/HE, and EpiAutoGP model tests
- Docker image build and deployment gate
- pre-commit and Codecov checks

Closes #1203


src/cfa/stf/routine/dagster_defs.py
```diff
@@ -353,6 +353,7 @@ def _run_fable_e_other(
         epiweekly=epiweekly,
         run_date=run_date,
         fail_on_stale_data=model_base_config.fail_on_stale_data,
+        logger=context.log,
```
giomrella, line 356: I would not redirect stdout/stderr into Dagster's context. The context events get sent to the Dagster database which can be slow and lead to memory issues besides causing general database bloat. By default, stdout/stderr are synced to Blob every 30 seconds which is much more scalable.

src/cfa/stf/routine/dagster_defs.py
```diff
@@ -401,6 +402,7 @@ def _run_pyrenew_model(
         rng_key=pyrenew_config.rng_key,
         run_date=run_date,
         fail_on_stale_data=model_base_config.fail_on_stale_data,
+        logger=context.log,
```
giomrella, line 405: (same comment as above repeated for targeting)

I would not redirect stdout/stderr into Dagster's context. The context events get sent to the Dagster database which can be slow and lead to memory issues besides causing general database bloat. By default, stdout/stderr are synced to Blob every 30 seconds which is much more scalable.

PR #1240, opened 2026-08-19
Simplify EpiAutoGP integration

### LINKED ISSUE #1239: Thoroughly review EpiautoGP module for any unnecessary code.

damonbayer:
Closes #1239.

## Summary

- build EpiAutoGP model input directly from the shared `ForecastRun` surveillance frames instead of serializing and rereading `combined_data.tsv`
- centralize NSSP daily-to-epiweekly aggregation in Python using `cfa.stf.forecasttools.daily_to_weekly` for both Fable serialization and EpiAutoGP input preparation
- pin the forecasttools change from [CDCgov/cfa-stf-forecasttools#234](https://github.com/CDCgov/cfa-stf-forecasttools/pull/234), which drops null and NaN values before strict week-completeness checks
- use Polars selectors to define NSSP value, index, and grouping columns
- remove the redundant R epiweekly conversion script and the now-unused `transform_serialized_data` pipeline hook
- preserve supported NSSP/NHSN target behavior
- replace generic parameter dictionaries with an explicit Julia runner interface
- remove the unused production `FixedNowcast` adapter and consolidate reporting-delay helpers
- replace stale module documentation with the current shared/model-specific boundary

## Why

EpiAutoGP retained infrastructure from before the repository introduced shared forecast-run state and pipeline orchestration. That left it reconstructing data already available in memory and maintaining abstractions that the production resolver never used.

Epiweekly NSSP aggregation was also implemented twice: once in an R rewrite of `combined_data.tsv` for Fable and once in EpiAutoGP-specific Polars code. Using the shared forecasttools implementation gives both models one aggregation path and removes an unnecessary serialization lifecycle step.

Forecasttools now removes invalid daily observations before aggregation. With `strict=True`, an affected series/week is incomplete and excluded, so this repository no longer needs a second weekly aggregation and join to detect nulls.

This reduces the EpiAutoGP and shared-pipeline surface while keeping model-specific responsibilities and output behavior intact.

## Validation

- `uv run pytest -m 'not model_integration and not pipeline_e2e' -q` — 157 passed, 4 deselected
- `just test-fable mock CA covid` — passed daily and epiweekly Fable configurations
- `just test-epiautogp mock CA covid` — passed all four EpiAutoGP configurations
- `uv run pre-commit run --files src/cfa/stf/routine/data/prep_data.py uv.lock` — passed
- `git diff --check` — passed


damonbayer:
Closed in favor of https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1255

Issue #1213, opened 2026-08-06
Run EpiAutoGP in Production


damonbayer:
## Preliminaries
- [x] https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1203
- [x] https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1239

## First Steps
- [ ] https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1292
	- with omitting a few days to avoid nowcasting
		- [x] https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/226)
	- As a designated model
	- Update Hubverse submitter
- [ ] Implement https://github.com/CDCgov/nowcastNHSN in production
- [ ] Submit EpiAutoGP-H in hubs
	- with https://github.com/CDCgov/nowcastNHSN for nowcasting
	- As a designated model
	- Update Hubverse submitter

## Model Next Steps
- [ ] Run EpiAutoGP-E-Denom and fuse with PyRenew-E
	- With truncation
- [ ] Run EpiAutoGP-Num with https://github.com/cdcent/cfa-parameter-estimates for nowcasting

## Technical Next Steps
- [ ] Determine appropriate VM to use with EpiAutoGP (may benefit from more cores)
	- @SamuelBrand1 suggest 64 cores would be good.
- [ ] Determine appropriate sampler parameters to use with EpiAutoGP
	- @SamuelBrand1 suggests:
	> number of mcmc structure moves per refresh: 200 seems ok
	>	Number of hmc moves per structure proposal: 50 seems to get there on default HMC settings
- [ ] Determine appropriate training period length to use with EpiAutoGP
	- Requires https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/841
	- More is better, but how does runtime scale?

## Longer Term
- [ ] Determine which E model should be used for internal E forecasts
- [ ] Determine if any other PyRnew/EpiAutoGP models should be designated
	- Can gather some anecdotal evidence as we start submitting
	- [ ] Deprecate Fable "other" models if not part of designation
- [ ] Should we be running an EpiAutoGP-E-daily model and performing model fusion, rather than EpiAutoGP-E-pct?

SamuelBrand1:
> More is better, but how does runtime scale?

The main runtime scaler is how many smc steps you make, so its a question of how much data between particle rejuvenations.

jkislin:
[...]

PR #1284, opened 2026-09-08
Allow different assets to use different Azure Batch VMs by default

### LINKED ISSUE #1270: Per-asset Azure VM's in Dagster Defs
In https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1213, we will want to run different models on differently sized Azure VM's. We can prepare for this now by updating our current dagster_defs to support this.

There is not a strong need for this right now, but we could probably run the Fable models on smaller VM's to save a bit of compute and prove the concept.

jkislin:
Seeks to address a piece of #1213 and #1270 in whole.
1. New execution configs that make use of the new `stf-routine-64cpu` and `stf-routine-2cpu` azure batch pools.
2. New automation condition sensors that submit runs against that config via run tags (we can also reference it in a schedule or asset job definition).
3. Reworked the expandable asset args accordingly; new and shorter asset group names.

#### Test - will an `eager()` asset inherit the run tags from the sensor?

The asset that needed the 64cpu pool as per the sensor got it:
image
and
image
and
image

Another asset inherited the default Azure Batch config:
image
(failure unrelated to this change - I forgot to push my latest image)

#### How to customize:
1. Make sure you have an Azure Batch pool pre-configured with the VM Size and configurations you'd like.
2. Define an Azure Batch Execution config accordingly. You can have multiple Azure Batch Execution configs.
3. Setup an Automation Condition Sensor, schedule, or job that specifies the Azure Batch Execution config you want.
4. Make use of `dg.AssetSelection` to link the desired model assets to the conditions, which are as mentioned linked to the config.
5. Test by making sure your asset lights up with the appropriate automation condition or schedule in the lineage page, or that you see the launchpad reflect your new pool on a job.

image
image


src/cfa/stf/routine/dagster_defs.py
```diff
@@ -572,6 +587,24 @@ def name(self) -> str:
...
+epiautogp_64cpu_sensor = dg.AutomationConditionSensorDefinition(
+    name="EpiAutoGP_64cpu",
+    target=dg.AssetSelection.groups("EpiAutoGP"),
+    run_tags=azure_batch_64cpu_execution_config.to_run_tags(),
+    default_condition=dg.AutomationCondition.eager(),
+    use_user_code_server=True,
+)
+
+
+@dg.asset(
```
jkislin: These dummy assets will be removed before merging.

src/cfa/stf/routine/dagster_defs.py
```diff
@@ -560,32 +585,62 @@ def name(self) -> str:
...
+    "group_name": "WeeklyFable",
+    "automation_condition": (
+        # We specifically don't want these to run unless it's Wednesday
+        # 0=monday,1=tuesday,2=wednesday,etc.
+        # Note this is different from cron which is 1-indexed
+        dg.AutomationCondition.eager() & IsWeekday(2)
+    ).with_label("eager_on_wed"),
+}
+
+weekly_pyrenew_args = {
```
damonbayer: Is there a way to tie `weekly_fable_args` `weekly_pyrenew_args` so we don't do the duplicative `dg.AutomationCondition.eager() & IsWeekday(2)` and `.with_label("eager_on_wed")`.

My mental dagster model is really straining here. This is getting quite complex.

jkislin: Yeah, we can actually put the automation condition in the base args, and downstream automation conditions will override it if we supply them. 

So this works just fine:
```python
weekly_base_args = {
    "partitions_def": daily_partitions_def,
    "retry_policy": dg.RetryPolicy(),  # allow the assets to retry once on failure
    "automation_condition": (
        # We specifically don't want these to run unless it's Wednesday
        # 0=monday,1=tuesday,2=wednesday,etc.
        # Note this is different from cron which is 1-indexed
        dg.AutomationCondition.eager() & IsWeekday(2)
    ).with_label("eager_on_wed"),
}

weekly_fusion_args = {
    **weekly_base_args,
    "group_name": "WeeklyFusion",
    # we want vanilla eager for the fusion assets
    "automation_condition": dg.AutomationCondition.eager(), # supersedes the earlier defined condition as its lower
}
```
Then, I'll just give `fable` and `pyrenew` assets the `weekly_base_args` and their appropriate `group_name` on the asset directly.

jkislin: I went even further simplifying this. Now, there is just `common_asset_args` and any overrides/specific-additions happen directly on the asset.

giomrella (approved):

damonbayer (approved):
Thanks @jkislin! Please update from `main` and you are good to merge.

PR #1289, opened 2026-09-09
Break `dagster_defs.py` into modules

### LINKED ISSUE #1271: Refactor dagster_defs
In https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1213, we are planning to add ~4 more model assets to our production workflow. https://github.com/CDCgov/cfa-stf-routine-forecasting/blob/main/src/cfa/stf/routine/dagster_defs.py is already quite long.

I think there are opportunities to split it up to make it easier to work with and also to take advantage newer developments in https://github.com/CDCgov/cfa-dagster, say by not having the redundant https://github.com/CDCgov/cfa-stf-routine-forecasting/blob/main/dagster_defs.py file.

One idea for simplification is that the dagster assets for each model might be better defined in the respective model folders in `src/cfa/stf/routine`.

jkislin:
> #1284 must be merged first

Instead of `uv run dagster_defs.py`, either call `cfa-dg dev` or `uv run dagster/defs.py`.

`dagster_defs.py` remains for now as a fallback. I will not be changing any functionality in this PR, only modularity/import structures.


jkislin:
Closing in favor of a slightly different approach.

PR #1297, opened 2026-09-10
Add epiweekly EpiAutoGP percentage asset

### LINKED ISSUE #1292: Add `epiautogp-e-pct-epiweekly` to dagster
We will add EpiAutoGP-E to our weekly dagster workflow.

`EpiAutoGP-E-pct` should

- fit directly to epiweekly NSSP data.
- Be added to the `e-data-rerun` tag
- Be a dependency of `postprocess_forecasts`
- Not be involved in any forecast fusion.
- use `n-forecast-draws = 2000`
- use `transformation = "percentage"`
- use `smc-data-proportion = 0.1`

In production, it should

- Run on the `azure_batch_64cpu_execution_config`
- Use `n-particles = 64`
- Use `n-mcmc = 200`
- Use `n-hmc = 50`
- Use all available training data from the comprehensive NSSP (pass `null` `n-lookback-days`)

In testing, it should

- Run on the `azure_batch_4cpu_execution_config`
- Use `n-particles = 4`
- Use `n-mcmc = 100`
- Use `n-hmc = 25`
- Use a lookback of 150 days data from the comprehensive NSSP.
- Be added to https://github.com/CDCgov/cfa-stf-routine-forecasting/blob/main/.github/workflows/containers.yaml

damonbayer:
## Summary

- add an `epiautogp_e_pct_epiweekly` asset backed by comprehensive epiweekly NSSP percentage data
- configure production runs for the 64-CPU pool with all available history and test runs for the 4-CPU pool with a 150-day lookback
- make the asset eligible for E-data reruns and a direct dependency of postprocessing without adding it to forecast fusion
- include the EpiAutoGP group in the container workflow's end-to-end asset job

Closes #1292

## Testing

- `uv run pytest -m "not pipeline_e2e and not model_integration"` (214 passed, 4 deselected)
- `uv run pytest tests/test_forecast_entrypoints.py tests/epiautogp/test_forecast_utils.py` (29 passed)
- `uv run ruff check src tests`
- `uv run ruff format --check src tests`
- `uv run dg check defs`

No Dagster jobs were launched.

### DIFF
--- a/src/cfa/stf/routine/dagster_defs.py
+++ b/src/cfa/stf/routine/dagster_defs.py
@@ -293,7 +293,7 @@ class ModelBaseConfig(_ModelTrainingFields, dg.ConfigurableResource):
         return self.model_copy(update=overrides)
 
 
-class FableEOtherConfig(dg.ConfigurableResource):  # used to inherit ModelBaseConfig
+class FableEOtherConfig(dg.ConfigurableResource):
     """
     Configuration for fable E-other model assets
     (fable_e_other, epiweekly_fable_e_other).
@@ -303,7 +303,7 @@ class FableEOtherConfig(dg.ConfigurableResource):  # used to inherit ModelBaseCo
     n_samples: int = 400 if not is_production else 2000
 
 
-class PyrenewConfig(dg.ConfigurableResource):  # used to inherit ModelBaseConfig
+class PyrenewConfig(dg.ConfigurableResource):
     """
     Configuration for Pyrenew model assets (pyrenew_e, pyrenew_h, pyrenew_he, etc.).
     These default values can be modified in the Dagster asset materialization launchpad.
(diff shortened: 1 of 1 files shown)


src/cfa/stf/routine/dagster_defs.py
```diff
@@ -315,6 +316,18 @@ class PyrenewConfig(dg.ConfigurableResource):  # used to inherit ModelBaseConfig
     additional_forecast_letters: str = ""
 
 
+class EpiAutoGPEPctEpiweeklyConfig(dg.ConfigurableResource):
+    """Configuration for the epiweekly EpiAutoGP E-pct model asset."""
+
+    n_lookback_days: int | None = None if is_production else 150
```
damonbayer, line 322: Do we need to remove `n_lookback_days` from `ModelBaseConfig`?

giomrella: The resources are distinct, so there is no risk of collision if that is your concern. I do notice that you are using the same default value here, so you could remove this `n_lookback_days` property if you don't need them to be distinct at runtime

giomrella (approved):

jkislin (approved):
I have personally tested this. Looks good.

src/cfa/stf/routine/dagster_defs.py
damonbayer: I want to default to `None` in production for EpiautoGP and `150` in testing. All the other models default to `150` in both contexts.

PR #1303, opened 2026-09-14
Clarify Dagster Launchpad configuration scope

### LINKED ISSUE #1302: Clarify which parameters are not inherited from the base config for epiautogp
We need to give these better names.

damonbayer:
## Summary

- preserve distinct default training windows: 150 days for Fable/PyRenew and all available history for EpiAutoGP in production
- expose one location-specific `n_lookback_days` setting in `config_overrides`; when explicitly supplied, it applies the same lookback to Fable, PyRenew, and EpiAutoGP
- distinguish an omitted override from explicit `null`: omission retains each model's default, while `null` makes every model use all available history for that location
- describe the scope of shared Launchpad fields and apply location-specific overrides consistently
- remove ineffective postprocessing copy controls from the Launchpad; postprocessing continues to always create the reorganized figure copy under the daily forecast output directory
- test configuration-schema and parameter routing behavior for Fable, PyRenew, EpiAutoGP, fusion output discovery, and postprocessing copy behavior

Closes #1302

## Testing

- `just test` (221 passed, 4 deselected)
- `uv run pytest tests/test_dagster_model_config.py` (8 passed)
- `uv run dg check defs`
- `uv run ruff check src/cfa/stf/routine/dagster_defs.py tests/test_dagster_model_config.py`
- `uv run ruff format --check src/cfa/stf/routine/dagster_defs.py tests/test_dagster_model_config.py`
- `git diff --check`


damonbayer:
@jkislin @giomrella Ready for your review. I'm not really sure about the quality / utility of the tests. Happy to remove or enhance them. They are just llm generated.

src/cfa/stf/routine/dagster_defs.py
```diff
@@ -289,6 +326,10 @@ def get_by_location(self, loc: Location) -> "ModelBaseConfig":  # type: ignore[r
                 entry = ConfigOverride(**entry)
             if entry.location == loc:
                 overrides = entry.model_dump(exclude={"location"}, exclude_unset=True)
+                if "n_lookback_days" in overrides:
+                    n_lookback_days = overrides.pop("n_lookback_days")
+                    overrides["fable_pyrenew_n_lookback_days"] = n_lookback_days
+                    overrides["epiautogp_n_lookback_days"] = n_lookback_days
```
damonbayer, line 332: I don't love this pattern, but I like what it enables. We want to have different defaults for different models. if you don't specifically override it, each model should use its default.

tests/test_dagster_model_config.py
```diff
@@ -0,0 +1,190 @@
...
+                fail_on_stale_data=False,
+            ).as_dict()
+        ],
+    )
+    config.diseases._current_value = "covid"
+    config.locations._current_value = "CA"
+    return config
+
+
+def test_launchpad_has_model_defaults_and_shared_location_override():
```
giomrella, line 32: I told Copilot that I thought these tests were low-value and it had this to say: 

> These assertions are just confirming that Pydantic field defaults work as Pydantic field defaults do—that's framework behavior, not business logic worth testing. The framework will automatically apply defaults; there's no custom code being validated here.
> 
> The genuinely valuable parts of this test are:
> 
> Lines 37-40: Asserting the absence of fields (n_lookback_days not in base, but IS in overrides) — this documents the schema design decision and prevents regression if someone accidentally adds fields to the wrong config class.
> 
> Lines 53-55: Checking that EpiAutoGPEPctEpiweeklyConfig doesn't expose n_lookback_days — this verifies separation of concerns at different config levels.
> 
> The default-value assertions (11-14, 46-52) could be removed without losing coverage. The test would be stronger if it focused purely on what's NOT there and where fields are intentionally missing, rather than confirming that frameworks set defaults correctly.
> 
> Good catch—that's a legitimate test smell worth refactoring out.

giomrella (approved):
I think the LLM added some low-value assertions to the new tests, but I think they are providing value overall by confirming the scope of various configuration values and the tiered defaults

Issue #1315, opened 2026-09-17
Workflow/Directory Structure Revamp


damonbayer:
This is a plan that should replace or make easier:
- https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1300
- https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1291
- https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1278
- https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/156.

## Current Status
The current directory structure is 

```
YYYY-MM-DD_forecasts/disease_lookback-x_omit-y/model_runs/state_abb/model_name
```

We have been using this structure because those are the two most common properties we modify when performing reruns. There are some issues with this:
1. `The lookback` and `omit` artificially limit the fusion forecasts that can be created. We expect models that are fusible to be contained within the same `disease_lookback-x_omit-y/model_runs/state_abb/` folder, but all models in `YYYY-MM-DD_forecasts` have the same `forecast_through` date, so they can be fused. (https://github.com/CDCgov/cfa-stf-routine-forecasting/issues/1291)
2. The `lookback` and `omit` properties more reflect the input parameters to a model, not that actual data used.
	1. Ex: models that don't perform nowcasting force a minimum omission of 3 days: when the user requests `omit=1`, it is overridden internally, but	the results are still saved to the `omit-1` directory (to comply with point 1).
	2. Ex:  Epiweekly models lookbacks and omissions reflect the daily data the epiweekly data is constructed from. So, to do fusion with a daily model with `omit=2`, we must fit a new epiweekly model with `omit=2` even though it has no impact on the actual data fit in the model (We need to do this to comply with point 1).
3. Performing a rerun where we do not change the `lookback` or `omit` (but do change something else) results in overwriting the previous fits.

## Proposal

I propose migrating to 

```
YYYY-MM-DD_forecasts/YYYY-MM-DD-HH-MM-SS/model_name/disease/state_abb/
```

Where the `YYYY-MM-DD-HH-MM-SS` is based on the launch time of the dagster job.

In general, the rest of the workflow (fusion and post-procesing) would be reworked to assume that a user wants to base everything off of the latest version of each named model. If that isn't the case, the user is free to manipulate the file system directly -- just delete or rename a result you don't want to use.

The launchpad yaml would be saved to 

```
YYYY-MM-DD_forecasts/YYYY-MM-DD-HH-MM-SS/
[...]

PR #1317, opened 2026-09-17
Refactor `dagster_defs.py` into a dagster project structure

jkislin:
### Summary
Addresses #1271 by splitting out `dagster_defs.py` into:
```bash
dagster/
├── definitions.py
└── defs
    ├── asset_config.py
    ├── asset_helpers.py
    ├── assets.py
    ├── automation.py
    ├── execution.py
    └── jobs.py
```

### Testing and further development
After pulling this branch, run `uv sync` and make sure your `.venv` is active.
Then, simply run `cfa-dg dev` to launch the dev webserver. 
- `dg dev` should also work, as should `uv run src/cfa/stf/routine/dagster/defs.py`

### Notes
- This has been tested with `cfa-dg check defs` and loading the UI locally without error. 
     - The UI experience should be identical.
- Numerous supporting files like `pyproject.toml` etc have been updated to reflect these changes.
- I used GPT-6 Astra to assist with the grunt work of refactoring/redistributing imports.


damonbayer:
@jkislin heads up that I have made several small changes to the dagster setup since this was opened
- https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1320
- https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1319
- https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1318

jkislin:
@damonbayer resolving conflicts by way of separate PR 
- #1321

jkislin:
Submitted a change to address the failing tests as `dagster_defs` is no longer a catch-all module to import.

src/cfa/stf/routine/dagster/definitions.py
```diff
@@ -0,0 +1,98 @@
...
+
+# Set Azure HTTP Logging Level
+# this will limit excessive IO logs in stderr
+# for any assets making azure http requests
+azure_http_logger = logging.getLogger(
+    "azure.core.pipeline.policies.http_logging_policy"
+)
+azure_http_logger.setLevel(logging.WARNING)
+
+collected_defs = collect_definitions(vars(assets) | vars(automation) | vars(jobs))
```
jkislin: @giomrella - curious on your thoughts here. 

We can no longer use `globals()` unless we either use wildcard imports (not good) or explicitly list every name (defeats the purpose). 

I've settled for now on `vars(<module>)`, which unlike wildcard imports keeps those modules' names from being dumped into the global scope.

Another alternative is to use a new dagster function in beta `dg.load_definitions_from_modules()`, but it warns of the possibility of breaking changes.

jkislin: See:
```python
 defs = dg.load_definitions_from_modules(
    modules=[assets, automation, jobs],
    resources={
        # These IOManagers let Dagster serialize asset outputs and store them
        # in Azure to pass between assets
        "io_manager": ADLS2PickleIOManager(),
        # Shared resources for model assets
        "model_base_config": asset_config.ModelBaseConfig(),
        "pyrenew_config": asset_config.PyrenewConfig(),
        "epiautogp_e_pct_epiweekly_config": asset_config.EpiAutoGPEPctEpiweeklyConfig(),
        "fable_e_other_config": asset_config.FableEOtherConfig(),
        "e_model_exclusions": asset_config.EModelExclusions(),
        "w_model_exclusions": asset_config.WModelExclusions(),
    },
    executor=dynamic_executor(
        default_config=execution.azure_batch_4cpu_execution_config,
        # default_config=execution.basic_execution_config,
        # default_config=execution.docker_execution_config,
        alternate_configs=[
            execution.basic_execution_config,
            execution.docker_execution_config,
        ],
    ),
)
```

jkislin:
@damonbayer @giomrella this PR is ready for review and ideally would be merged before any other changes are made to `dagster_defs.py`, which is fully removed here.

src/cfa/stf/routine/dagster/asset_config.py
```diff
@@ -0,0 +1,185 @@
...
+
+import dagster as dg
+from cfa.stf.forecasttools import LOCATION_LIST
+from cfa_dagster import GraphDimension, GraphDimensionExclusion
+from cfa_dagster import is_production as is_prod
+from pydantic import BaseModel, Field
+
+# ============================================================================
+# GRAPH DIMENSIONS AND PARTITIONS
+# How are the data split and processed in Azure Batch?
```
damonbayer: Not data, but… jobs? runs? tasks?

src/cfa/stf/routine/dagster/asset_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+from cfa.stf.routine.pyrenew_hew.forecast_pyrenew import main as forecast_pyrenew
+from cfa.stf.routine.utils.prop_utils import create_prop_fusion_model
+from cfa.stf.routine.utils.r_utils import (
+    make_figures_from_model_fit_dir,
+    model_fit_dir_to_hub_tbl,
+)
+
+# ============================================================================
+# MODEL CONSTRUCTOR FUNCTIONS - these are used later, in Asset Definitions
+# ============================================================================
```
damonbayer:
```suggestion
```

Now self explanatory, based on `asset_helpers` file name

src/cfa/stf/routine/dagster/asset_helpers.py
```diff
@@ -0,0 +1,258 @@
...
+        smc_data_proportion=epiautogp_e_pct_epiweekly_config.smc_data_proportion,
+        n_threads=epiautogp_e_pct_epiweekly_config.n_threads,
+        nowcast_source_name="none",
+        run_date=run_date,
+        fail_on_stale_data=loc_config.fail_on_stale_data,
+        logger=context.log,
+    )
+
+
+def _get_model_loc_dir(
```
damonbayer: Move this above `_run_fable_e_other`

src/cfa/stf/routine/dagster/assets.py
```diff
@@ -0,0 +1,326 @@
...
+    "e-data-rerun-no-epiautogp": "",
+}
+EPIAUTOGP_E_DATA_RERUN_TAGS = {"e-data-rerun": ""}
+H_DATA_RERUN_TAGS = {"h-data-rerun": ""}
+HE_DATA_RERUN_TAGS = E_DATA_RERUN_TAGS | H_DATA_RERUN_TAGS
+
+# ============================================================================
+# ASSET DEFINITIONS
+# ============================================================================
+# These are the core of Dagster - functions that specify data
```
damonbayer:
```suggestion
```

src/cfa/stf/routine/dagster/defs/assets.py
```diff
@@ -0,0 +1,326 @@
...
+    "e-data-rerun": "",
+    "e-data-rerun-no-epiautogp": "",
+}
+EPIAUTOGP_E_DATA_RERUN_TAGS = {"e-data-rerun": ""}
+H_DATA_RERUN_TAGS = {"h-data-rerun": ""}
+HE_DATA_RERUN_TAGS = E_DATA_RERUN_TAGS | H_DATA_RERUN_TAGS
+
+# ============================================================================
+# ASSET DEFINITIONS
+# ============================================================================
```
damonbayer, line 50: Can you take a look over all comment decorations? Let's be sure they're still useful now that the files are split up.

I also think I'd like to limit to 

```python
# Regular comments
```

and

```python
# =============================================================================
# Section Breaks
# =============================================================================
```
 but let's get rid of the non-centered 

```python
# ----- These things -----

File: src/cfa/stf/routine/dagster/assets.py [hunk 7]
```diff
@@ -0,0 +1,326 @@
...
+):
+    _fuse_pyrenew_fable_e_other(
+        context,
+        model_base_config,
+        pyrenew_model_name="pyrenew_he",
+        epiweekly=True,
+    )
+
+
+# ---------- Postprocessing Forecast Batches ----------
```
damonbayer:
```suggestion
# ---------- Postprocessing ----------
```

src/cfa/stf/routine/dagster/automation.py
```diff
@@ -0,0 +1,97 @@
...
+        ]
+        return f"is_{days[self.weekday].lower()}"
+
+
+eager_on_wed = (
+    # We specifically don't want these to run unless it's Wednesday
+    # 0=monday,1=tuesday,2=wednesday,etc.
+    # Note this is different from cron which is 1-indexed
+    dg.AutomationCondition.eager() & IsWeekday(2)
+).with_label("eager_on_wed")
```
damonbayer: I think could be made more clear by importing calendar from the standard lib.

Codex suggests

```python
from calendar import Day
from zoneinfo import ZoneInfo

NEW_YORK = ZoneInfo("America/New_York")


class IsWeekday(dg.AutomationCondition):
    def __init__(self, weekday: Day):
        self.weekday = weekday

    def evaluate(self, context: dg.AutomationContext) -> dg.AutomationResult:
        evaluation_time = context.evaluation_time.astimezone(NEW_YORK)

        return dg.AutomationResult(
            context=context,
            true_subset=(
                context.candidate_subset
                if evaluation_time.weekday() == self.weekday
                else context.get_empty_subset()
            ),
        )

    @property
    def name(self) -> str:
        return f"is_{self.weekday.name.lower()}"


eager_on_wednesday = (
    dg.AutomationCondition.eager()
    & IsWeekday(Day.WEDNESDAY)
).with_label("eager_on_wednesday")
```

src/cfa/stf/routine/dagster/automation.py
```diff
@@ -0,0 +1,97 @@
...
+    name="Postprocess",
+    target=dg.AssetSelection.groups("Postprocess"),
+    run_tags=azure_batch_2cpu_execution_config.to_run_tags(),
+    use_user_code_server=True,  # allows for custom automation conditions
+)
+
+epiautogp_sensor = dg.AutomationConditionSensorDefinition(
+    name="EpiAutoGP",
+    # add a group_name="EpiAutoGP" to an epiautogp asset to include it
+    # in the rules and configuration this sensor provides
```
damonbayer:
```suggestion
```

uv.lock
damonbayer, line 1: These don't seem relevant?

src/cfa/stf/routine/README.md
```diff
@@ -171,7 +171,7 @@ EpiAutoGP also requires at least three when it runs without a nowcast source.
 ## 4. Add orchestration explicitly
 
 Pipeline subclasses are not discovered automatically.
-To run the new model in production, update [`dagster_defs.py`](dagster_defs.py) to:
+To run the new model in production, update [`dagster/defs.py`](./src/cfa/stf/routine/dagster/defs.py) to:
```
damonbayer: Check all the links. This one resolves to `src/cfa/stf/routine/src/cfa/stf/routine/dagster/defs.py` when it should go to `src/cfa/stf/routine/dagster/defs.py`

damonbayer (changes requested):
Didn't complete a full review yet, but I left some comments to address.

src/cfa/stf/routine/dagster/asset_config.py
jkislin: Actually, we could make "data" true as well if we want to re: #1315.

src/cfa/stf/routine/dagster/definitions.py
giomrella: It seems like we're pretty close to Dagster's recommended approach if we rename `defs.py` to `definitions.py` and create a `dagster/defs/` folder. Let me take a quick crack at it.

```python
# definitions.py
from pathlib import Path

from dagster import definitions, load_from_defs_folder


@definitions
def defs():
    return load_from_defs_folder(project_root=Path(__file__).parent.parent.parent)
```

jkislin: I played with that. You don't need to name the file `definitions.py` nor do you need to name the folder `defs` and keep the definitions file out of it. You just need to configure `pyproject.toml` such that it knows where to look for your modules. I personally find having both a `defs` folder and a `definitions.py` to be very unintuitive naming wise. It gives me the ick, as they say.

The only reason we'd want to use the exact structure they give in the tutorial, from what I can see, is if we'd want the default behavior of `dg scaffold` to work properly. I'm not too concerned with that since this project is already "scaffolded"

jkislin: ~That said, `load_from_defs_folder()` might be a reasonable option. We just have to specify the `project_root` correctly.~

`load_from_defs_folder()` is mostly in service of building components. I'm not sold on it unless we go that route.

https://docs.dagster.io/api/dagster/components

src/cfa/stf/routine/dagster/asset_config.py
damonbayer: I guess it depends if we are talking about input data or output data!

src/cfa/stf/routine/dagster/definitions.py
jkislin: Dagster seems to be working on some good options to enhance flexibility and auto-detection:
https://docs.dagster.io/api/dagster/definitions

giomrella: @jkislin @damonbayer What do you think about this structure? https://github.com/CDCgov/cfa-stf-routine-forecasting/pull/1324/changes

src/cfa/stf/routine/dagster/automation.py
giomrella: I never liked `isWeekday(int)`. I definitely vote for using the standard, clearly-named enum in a future PR

src/cfa/stf/routine/dagster/asset_config.py
jkislin:
```
# ============================================================================
# GRAPH DIMENSIONS AND PARTITIONS
# i.e. "How is execution parallelized?"
# ============================================================================
```

In general I am thinking to use this template:
```
# ============================================================================
# ONTOLOGIES COVERED IN SECTION
# i.e. "Question developer or operator would want to answer?"
# ============================================================================
```

uv.lock
jkislin: These are the result of cfa-dagster updates. I've just pushed another.

src/cfa/stf/routine/dagster/automation.py
jkislin: `isWeekday(int)` is necessary because the dagster automation context encodes weekdays as such. 
I could make a mapping, but then we're just offloading this problem to a different piece of the code.

jkislin: Specifically, `context.evaluation_time.astimezone(NEW_YORK).weekday()` returns a value with type `int`

jkislin: Strike this... `Day.MONDAY == 0` is true, Tuesday is 1, etc. :facepalm:

jkislin: Pushing change.

giomrella: I think Damon's example code here would "just work" because the `calendar.Day.<WEEKDAY>.value` is also an int and happens to use the same values as the underlying `datetime.weekday()`

src/cfa/stf/routine/dagster/definitions.py
giomrella: @jkislin I think the solution you have in this PR is OK to merge, though I'd rather go with Dagster's recommended file structure if we're going to show people this repo as the blueprint for modularization. If Dagster releases more module loading functions out of preview/beta, I'd prefer to deprecate and later remove the custom `collect_definitions()` function.

jkislin:
@giomrella after digging around more through the dagster docs and testing your branch, I'm no longer as strongly opposed. We can play around with it.

```
dagster/
├── definitions.py
└── defs
    ├── asset_config.py
    ├── asset_helpers.py
    ├── assets.py
    ├── automation.py
    ├── execution.py
    └── jobs.py
```
Updated my original summary here to reflect this.

This is now ready for re-review. Thanks all!

pyproject.toml
```diff
@@ -51,9 +51,8 @@ source-exclude = [
 directory_type = "project"
 
 [tool.dg.project]
-root_module = "cfa.stf.routine"
-defs_module = "cfa.stf.routine.dagster_defs"
-code_location_target_module = "cfa.stf.routine.dagster_defs"
+root_module = "cfa.stf.routine.dagster"
+defs_module = "cfa.stf.routine.dagster.defs"
```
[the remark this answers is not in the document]
giomrella: Confirmed locally, you get an error unless you remove the `defs_module`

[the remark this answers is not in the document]
jkislin: Done.

src/cfa/stf/routine/dagster/defs/execution.py
```diff
@@ -0,0 +1,179 @@
...
+        config={
+            "image": image,
+            "retries": {"enabled": {}},
+            "container_kwargs": {
+                "volumes": [
+                    # bind the ~/.azure folder for optional cli login
+                    f"/home/{user}/.azure:/root/.azure",
+                    # bind current file so we don't have to rebuild
+                    # the container image for workflow changes
+                    f"{local_workdir}:{container_workdir / 'src/cfa/stf/routine/dagster'}",
```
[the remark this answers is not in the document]
jkislin: Done.

jkislin:
All tests (sans codecov) are passing and all existing review comments have been addressed.

Additionally, everything loads on the prod server:
image
And my test run completed without error:
image

giomrella (approved):

PR #1322, opened 2026-09-17
Have sensors turn on by default in production

jkislin:
A simple template for ensuring any new or changed sensors have their default status turned on without manual intervention.

This should not change the status of any existing sensors in any direction.

```
default_status=(
        dg.DefaultSensorStatus.RUNNING if is_prod() else dg.DefaultSensorStatus.STOPPED
),
```


jkislin:
@damonbayer this makes sense to merge into main directly after #1317
I am leaving it as a draft until that one is merged prior.

PR #1324, opened 2026-09-18
using dagster recommended project structure

giomrella:
...so we can use `load_from_defs_folder()`

This successfully loads the definitions with `cfa-dg dev`. I made a small change in `cfa-dagster` to suppress an inconsequential warning about mismatched code location names. Untested in docker/batch/as a code location


jkislin:
Hold on this for a second, I'm trying something that might be even simpler than either approach.

jkislin:
I see the utility in using exactly what dagster has in their tutorial (ease of adoption from/to vanilla dagster, `dg scaffold`, etc). Additionally, one benefit we do get with `load_defs_from_folder()` is that you can make a dagster object in any python file in the `defs/` folder and it will find it.  

However, my vote is against this for the following reasons:
- The dagster docs present various template structures - the `defs/` folder structure is just the structure recommended by the beginner tutorial.
- The semantic similarity between `definitions.py` and `defs/` is confusing.
- My approach emulates the approach you'd use with the beta function `dg.load_definitions_from_modules()`, which doesn't at all require the tutorial-recommended structure. It would be a quick swap in the future.
- The tutorial-recommended structure's complexity is what drove us to create a single `dagster_defs.py` file in the first place.

jkislin:
Playing around with this, I see the benefit of sticking with dagster's tutorial structure. I definitely want to revisit this at some point though.

